PR #1, opened 2016-07-29
Correct checkout command

curiousleo:
- Not everyone has the `co` alias set up
- Fix the branch name


mwitkow:
readme changed massively.

PR #24, opened 2017-04-24
Errors through Codes and using BrowserStack for testing

MarcusLongmuir:
This PR removes the `onError` callback in favour of reporting errors through gRPC codes.

It also switches the browser testing to BrowserStack from SauceLabs. This was necessary to test CORS handling.


ts/src/transports/Transport.ts
```diff
@@ -19,6 +21,31 @@ export type TransportOptions = {
...
 }
 
+let xhr: XMLHttpRequest;
+function getXHR () {
+  if (xhr !== undefined) return xhr;
+
+  if (XMLHttpRequest) {
+    xhr = new XMLHttpRequest();
+    try {
+      xhr.open('GET', 'https://example.com')
```
mwitkow: this is weird, why is this needed?

ts/src/transports/Transport.ts
```diff
@@ -33,6 +60,10 @@ export class DefaultTransportFactory {
       return fetchRequest;
     }
 
+    if (xhrSupportsResponseType("ms-stream")) {
```
mwitkow: add a comment about this mumbo jumpbo.

ts/src/transports/msStream.ts
```diff
@@ -0,0 +1,60 @@
+import {BrowserHeaders} from "browser-headers";
```
mwitkow: Can you add a comment to this whole file as to why this magic is needed + add the doc links to where you found these incantations? :)

ts/src/grpc.ts
```diff
@@ -130,53 +171,62 @@ export namespace grpc {
...
       onChunk: (chunkBytes: Uint8Array) => {
-        props.debug && console.debug("onChunk: ", chunkBytes);
-
         let data: Chunk[] = [];
         try {
           data = parser.parse(chunkBytes);
         } catch (e) {
-          props.onError(e);
+          props.debug && debug("onChunk.parsing error", e, e.message);
+          rawOnError(Code.Internal, e.message);
```
mwitkow: Ok, so these are parsing errors. Can you add a "parsing error" prefix to the message?

ts/src/grpc.ts
```diff
@@ -130,53 +171,62 @@ export namespace grpc {
...
       },
       onComplete: () => {
-        props.debug && console.debug("grpc.onComplete");
+        props.debug && debug("grpc.onComplete");
 
         if (responseTrailers === undefined) {
           if (responseHeaders === undefined) {
             // The request was unsuccessful - it did not receive any headers
-            rawOnError(new Error("Response closed without grpc-status (No headers)"));
+            rawOnError(Code.Unknown, "");
```
mwitkow: can you add text back to the error?

ts/src/grpc.ts
```diff
@@ -130,53 +171,62 @@ export namespace grpc {
...
+            rawOnMessage(deserialized);
           } else if (d.chunkType === ChunkType.TRAILERS) {
+            props.debug && debug("onChunk.trailers", responseTrailers);
             responseTrailers = new BrowserHeaders(d.trailers);
           }
         });
       },
       onComplete: () => {
-        props.debug && console.debug("grpc.onComplete");
+        props.debug && debug("grpc.onComplete");
```
mwitkow: what happens on resolution failures etc? is that gonna go through onComplete? If so, can you make sure we return a proper error here? and maybe add a comment as to why :)

ts/src/transports/Transport.ts
MarcusLongmuir: This can be `https://localhost`. The transport feature detection requires an `XMLHttpRequest`, which can only be constructed with a valid URL. It doesn't actually make a request.

ts/src/transports/Transport.ts
MarcusLongmuir: Will add in the `ms-stream.ts` file

mwitkow:
LGTM 🚢

PR #28, opened 2017-04-27
Follow lint suggestions

matiasinsaurralde:


easyCZ:
Will close as this is unlikely to get merged given the current shape of the proxy.

PR #35, opened 2017-05-02
Add +proto to grpc-web content-type

anuraaga:
My reading of the wire format for both web and native seem to indicate the ```+proto```, etc component is required. As grpc-web currently is tied to proto, it seems safe to just fix the content type to +proto for now.

https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md
http://www.grpc.io/docs/guides/wire.html

### DIFF
--- a/ts/src/grpc.ts
+++ b/ts/src/grpc.ts
@@ -171,7 +171,7 @@ export namespace grpc {
 
   export function invoke<TRequest extends jspb.Message, TResponse extends jspb.Message, M extends MethodDefinition<TRequest, TResponse>>(methodDescriptor: M, props: RpcOptions<TRequest, TResponse>) {
     const requestHeaders = new BrowserHeaders(props.metadata ? props.metadata : {});
-    requestHeaders.set("content-type", "application/grpc-web");
+    requestHeaders.set("content-type", "application/grpc-web+proto");
     requestHeaders.set("x-grpc-web", "1"); // Required for CORS handling
 
     const framedRequest = frameRequest(props.request);


mwitkow:
@anuraaga do you have any particular reason for this change?

Currently most (all?) gRPC implementations have only one single codec that they can run with. As such the `+proto` subtype negotiation is not really supported at the moment. For example, here's a lengthy PR in grpc-go that is probably gonna get stuck for a wider discussion in grpc mailing list:
https://github.com/grpc/grpc-go/pull/1203

anuraaga:
Hi @mwitkow - for context, I come from Java and am using both the official GRPC clients as well as my own custom Java GRPC server (https://github.com/line/armeria/tree/master/grpc) - it's already set up for content-type negotation and will support both proto and JSON. I ran into this issue because that implementation currently requires `+proto` so didn't accept a grpc-web request properly.

While most servers don't seem to do any actual content-type negotation and thus accept non-qualified `application/grpc`, I know at least the Java client does set `+proto` and would expect that many do (does Go not set it?). I don't see any reason not to provide a precise content-type if it's possible and simple.

That said, I re-read the GRPC Wire Protocol document, and realized it doesn't seem to be clear on the specification of the content type. Content-Type is required (https://github.com/grpc/grpc-java/blob/master/netty/src/main/java/io/grpc/netty/NettyServerHandler.java#L599) yet is in brackets in the doc, which confused me into thinking the `+proto` is also required. I guess there's a typo that Content-Type should have not been in brackets, and indeed the `+proto` may be optional. It doesn't specify what the format is when not specified though...

I'd understand not merging this PR since the spec isn't clear - at the same time, again I don't see any reason to not provide the precise content-type, so would still lean towards having it for robustness (thrift RPC is a mess when it comes to content-type negotiation since many clients use the same unspecified content-type with different serialization formats, ugh...).

mwitkow:
@wenbozhu can you comment on the `content-type` negotiation? I think it would be safe to make it default to `application/grpc-web+proto` but I'd like to know what the wider picture is :)

wenbozhu:
@mwitkow  Yes, I think it's a good idea to specify +proto with the sender; while receiver assumes proto being the default message format.

I'll update the spec doc.

wenbozhu:
https://github.com/grpc/grpc/pull/11236

mwitkow:
Cool!
Thanks @anuraaga,@wenbozhu!

anuraaga:
Thanks for confirming! @mwitkow just wondering, did you mean to close this PR without merging?

mwitkow:
Clicking is hard :D

PR #38, opened 2017-05-04
Added unary convenience function and header/trailer combination tests

MarcusLongmuir:


ts/src/transports/Transport.ts
```diff
@@ -64,10 +71,14 @@ export class DefaultTransportFactory {
       return mozXhrRequest;
     }
 
+    if (XMLHttpRequest.prototype.hasOwnProperty("overrideMimeType")) {
```
MarcusLongmuir: Preferring XHR with overrideMimeType over ms-stream due to 16MB limitation with ms-stream

ts/src/grpc.ts
```diff
@@ -102,17 +124,46 @@ export namespace grpc {
...
-        return Code.Internal;
+        return null;
       }
     }
-    return Code.Internal;
+    return null;
+  }
+
+  export function unary<TRequest extends jspb.Message, TResponse extends jspb.Message, M extends UnaryMethodDefinition<TRequest, TResponse>>(methodDescriptor: M,
+                                                                                                                                        props: UnaryRpcOptions<M, TRequest, TResponse>) {
```
jonny-improbable: random empty line.

MarcusLongmuir: It's the formatting. The line has the rest of the parameters on, but they're off-screen on GH. I'll just put them all on the line above.

ts/src/grpc.ts
```diff
@@ -102,17 +124,46 @@ export namespace grpc {
...
       } catch (e) {
-        return Code.Internal;
+        return null;
       }
     }
-    return Code.Internal;
+    return null;
+  }
+
+  export function unary<TRequest extends jspb.Message, TResponse extends jspb.Message, M extends UnaryMethodDefinition<TRequest, TResponse>>(methodDescriptor: M,
```
jonny-improbable: What happens if the client makes a `unary()` call against a streaming endpoint?  If the result is predictable (ie: just the first message in the stream) then that would make sense, however if it's not predictable then I would suggest we throw a runtime exception to avoid suprises.

ts/src/transports/Transport.ts
jonny-improbable: Might just be worth dropping ms-stream entirely?

ts/src/grpc.ts
MarcusLongmuir: Will add a runtime exception if used on a streaming method 👍

ts/src/transports/Transport.ts
MarcusLongmuir: It's the only transport for IE 10 and 11 due to their lack of `overrideMimeType`

jonny-improbable: 😱

Issue #53, opened 2017-06-02
Can the  grpc-web-client run on node?


HTChang:
Our website use the Universal Rendering, I succeeded to run the standalone grpcwebproxy with client side rendering, but when it comes to server side rendering, I got the error
```
ReferenceError: XMLHttpRequest is not defined
     at getXHR (/**/node_modules/grpc-web-client/src/transports/Transport.ts:27:7)
     at xhrSupportsResponseType (/**/node_modules/grpc-web-client/src/transports/Transport.ts:37:15)
     at Function.DefaultTransportFactory.detectTransport (/**/node_modules/grpc-web-client/src/transports/Transport.ts:62:9)
     at Function.DefaultTransportFactory.getTransport (/**/node_modules/grpc-web-client/src/transports/Transport.ts:52:47)
     at Object.invoke (/**/node_modules/grpc-web-client/src/grpc.ts:160:43)
     at /**/src/helpers/books.js:11:10
     at F (/**/node_modules/core-js/library/modules/_export.js:35:28)
     at batchGetBook (/**/src/helpers/books.js:10:10)
     at /**/src/redux/modules/full.js:96:12
     at Object.promise (/**/src/containers/Book/Book.js:17:19)
     at /**/node_modules/redux-async-connect/lib/ReduxAsyncConnect.js:118:34
     at Array.reduce (native)
     at /**/node_modules/redux-async-connect/lib/ReduxAsyncConnect.js:117:35
     at Array.map (native)
     at loadAsyncConnect (/**/node_modules/redux-async-connect/lib/ReduxAsyncConnect.js:114:68)
     at loadOnServer (/**/node_modules/redux-async-connect/lib/ReduxAsyncConnect.js:142:16)
```
it's because `XMLHttpRequest` is a built-in object in web browsers but not distributed with Node.
Are there any chance to let the grpc-web works on the Universal Rendering pattern of our website?

MarcusLongmuir:
@HTChang: I've open a PR for Node.js support (see link above). Can you please verify that this version suits your needs?

You should be able to install this branch using `"grpc-web-client": "git://github.com/improbable-eng/grpc-web.git#feature\/node-support"` in your `package.json`.

PR #55, opened 2017-06-11
Node.js support

MarcusLongmuir:
I've added support for Node.js using Node's `http` and `https` modules as initially requested in #53.

This required a refactor of test setup and addition of the `text-encoding` package.

The library is currently only tested with Webpack (and now Node.js). It's possible this change has broken other environments due to one of the transports requiring node-only modules.

@HTChang, can you confirm that this approach works?


mwitkow:
I am not a node.js guru, but technically this looks fine.

Please add a readme before the merge that states clearly that this is an unofficial nodejs api.

PR #56, opened 2017-06-13
Add cpp example

osechet:
This PR adds a cpp example to work with the proxy. The README in `example/cpp/` explains how to use it.


MarcusLongmuir:
Hey @osechet. Thanks for raising. For the time being we're focussing on Golang and TypeScript, but if you create a new repo for this I'm happy to link to it in the readme.

PR #58, opened 2017-06-21
Send cookies to same origin with fetch transport to have same behavio…

anuraaga:
…r among all transports.

While it might be useful to have an option, it would be a bit weird since the option would only have an effect on the fetch transport, not xhr transports. Let me know if an option would still be preferable.

While most gRPC APIs themselves shouldn't rely on cookies for state per good API design, it's still useful to send them from browsers since many proxies rely on them for things like corp-authentication (I'm using GCP's Identity-Aware Proxy) and sticky sessions.


MarcusLongmuir:
Thanks for the fix

PR #76, opened 2017-08-17
Fix small error in README

crisbodnar:
- if statement was missing ()
- grpc.Code does not exist. Code.OK should be used instead


ts/README.md
```diff
@@ -42,7 +42,7 @@ grpc.invoke(BookService.QueryBooks, {
     console.log("got book: ", message.toObject());
   },
   onEnd: (code: grpc.Code, msg: string | undefined, trailers: Metadata) => {
```
MarcusLongmuir: `grpc.Code` on this line should also be changed

PR #78, opened 2017-08-17
Add support for aborting RPC requests.

### LINKED ISSUE #66: Unclear how to stop a client.invoke on a stream
I have a working example of accessing a stream using `grpc.invoke` however there seems to be no way to cancel.  Would think there would be something like `setInterval` ala...

```
const cancelId = grpc.invoke(...)
grpc.cancelnvoke(cancelId)
```

thoughts?

jonnyreeves:
This diff requests `Transport` instances to return a `CancelFunc` which consumers can invoke to abort the request initiated by the `Transport`.

`grpc.invoke` and `grpc.unary` both return a `Client` object which exposes an `abort` method which can be used to invoke the `Transport`'s `CancelFunc` thereby aborting the request.

Fixes #66


test/hosts-config.ts
```diff
@@ -1,3 +1,3 @@
 // These must match the hosts configured in .travis.yml
-export const testHost = "testhost";
-export const corsHost = "corshost";
+export const testHost = "localhost";
+export const corsHost = "localhost";
```
MarcusLongmuir: Unfortunately this is necessary to cause IE to treat the different addresses (different ports) as different origins so that CORS behaviour can be tested. IE treats `localhost:1234` and `localhost:5678` as the same origin 😕

Relevant test failure: https://travis-ci.org/improbable-eng/grpc-web/builds/265692040#L1210

jonnyreeves:
Ah sorry this was me debugging (I was having a problem with getting chrome to accept my self signed cert). I'll revert this change

PR #81, opened 2017-08-26
Change wrapped server to a http.Handler

johanbrandhorst:
This allows the wrapped server to be used in a composable manner with other Go handlers.

### DIFF
--- a/go/grpcweb/helpers.go
+++ b/go/grpcweb/helpers.go
@@ -5,8 +5,6 @@ package grpcweb
 
 import (
 	"fmt"
-
-	"google.golang.org/grpc"
 )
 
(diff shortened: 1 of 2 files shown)


johanbrandhorst:
Hm, this was less trivial than I expected.

Basically, my usecase is thus: I want to create a `http.Handler` that conditionally calls other handlers based on the requests. I would like to include the grpcproxy in that chain, such that I can have the grpcproxy wrap a handler of mine that conditionally switches between calling to a gRPC server handler or a `http.FileServer` based on headers. Unfortunately the `ListGRPCResources` function requires a `grpc.Server`.

I suppose I could define an interface which would still make this kind of chaining possible, but I'd understand if that change is to large.

johanbrandhorst:
This is old and unloved so I'm closing it.

PR #86, opened 2017-09-10
cache registered service endpoints for more efficient routing

pkieltyka:
`ListGRPCResources()` is called for each request to the server, which is unnecessary. Instead we save the slice upon creation of the server.


easyCZ:
Rebased.

PR #88, opened 2017-09-14
Add .abort() to requests

### LINKED ISSUE #66: Unclear how to stop a client.invoke on a stream

MarcusLongmuir:
This PR adds a return object from `invoke` and `unary` that contains a single `abort: () => void` property that can be called to abort the request. This abort signal is passed to the transport, which then closes the connection to the server.

I've refactored the streaming tests to avoid using time, which made them flaky, into using a side channel that is used to request the next message be sent by the server.

This pattern allowed writing a much cleaner test for cancellation, but it still requires polling for a few seconds after sending the abort because some browsers (Edge 14 of those that were tested) don't close the connection immediately.

This fixes #66 

Credit to @jonnyreeves for the actual aborting code.


midan888:
Guys, are you gone fix and merge this PR ? 
Nice work, I really need abort functionality for streams ASAP, thanks for what you are doing here!

easyCZ:
Good work @MarcusLongmuir and @jonnyreeves. I think we should also call `onEnd()` with a CANCELLED status once a request is cancelled. This may, however, be a breaking change as existing users may be ignoring the status returned by `onEnd()`. What do you think?

MarcusLongmuir:
@easyCZ: It seems that the status code you linked indicates that it should be, so invoking `onEnd` with `CANCELLED` makes sense. Do you know if this is the case in any other implementations?

jonny-improbable:
I'd vote to land this as-is and consider ending the a `CANCELLED` status in a subsequent PR as it would be a non breaking change.

MarcusLongmuir:
This has been released in `grpc-web-client@0.3.1`

midan888:
@MarcusLongmuir cool, thank a lot!

PR #90, opened 2017-09-22
Add grpcweb to get_go_deps.sh

telekid:
Running `npm start` in the example project failed because `grpcweb` had never been installed into my go workspace. This pull request will fix that issue for future users.


easyCZ:
Thanks for your PR, we will not be merging this since #113 Resolves this issue by using `dep`

Issue #87, opened 2017-09-11
Debugger Interface Proposal


easyCZ:
This is a proposal for a debugger interface in grpc-web-client. The intent is to make the current console based debugger extensible and allow clients to implement their own version of the debugger. The driving factor is to improve the currently tedious debugging experience where grpc-web requests shown in the Network tab are not properly serialized and hard to inspect. The proposed interface allows for an extension to be build for Chrome & Firefox to expose grpc-web requests similarly to the Network tab in Chrome.

This document is an aimed athelping shape the debugger interface and get feedback on the proposed changes.

## Interface
The interface proposed consists of a `DebuggerProvider` and a `Debugger`.

### DebuggerProvider
The provider is responsible for supplying an instance of a debugger to be used in a particular request. Additionally, it can implement supplemental functionality such as tracking of requests or recording request history.
```typescript
export interface DebuggerProvider {
    // Obtain an instance of a debugger.
    // A convenience proxy allowing the provider to implement additional functionality
    // such as tracking history of requests
    getInstanceForRequest(id: number): Debugger;
}
```

### Debugger
The `Debugger` implements callback methods for the lifecycle of a grpc-web request.
```typescript
// Implements callbacks for grpc-web request/response lifecycle events
export interface Debugger {
    // Called just before a request is fired, called only once
    onRequestStart(host: string, method: MethodDefinition): void;

    // Called just before a request is fired, called only once
    onRequestHeaders(headers: BrowserHeaders): void;

    // Called with the request payload, potentially called multiple times with request streams
    onRequestMessage(payload: Message): void;

    // Called when response headers are available, called once multiple times
    // This is a low level method intended to debug byte serialization
    onResponseHeaders(headers: BrowserHeaders, httpStatus: number): void;

    // Called with each received chunk
    onResponseChunk?(chunk: Chunk[], chunkBytes: Uint8Array): void;

    // Called with each response message, called multiple times with response streams
    onResponseMessage(payload: Message): void;

    // Called with response trailers, called once
    onResponseTrailers(metadata: BrowserHeaders): void;

    // Called when a request completes, called once
    onResponseEnd(grpcStatus: Code | null): void;

    // Called with any error occuring in the flow, potentially called multiple times
    onError(code: Code, err: Error): void;
}
```

### Example
```typescript
export class ConsoleDebuggerProvider implements DebuggerProvider {

    getInstanceForRequest(id: number): Debugger {
        return new ConsoleDebugger(id);
    }

}

[...]

PR #95, opened 2017-10-16
Debugger Interface

easyCZ:
Adds debugger interface support.

Implements the following:
* Ability to register a debugger, deregister and get the current set of debuggers
* DebuggerFactory interface
* Debugger interface
* ConsoleDebugger implementation
* DebuggerDispatch to help dispatch to all registered debuggers

This is an implementation of #87.

Resolution of deprecated fields is still work in progress. This is for a first pass review.


easyCZ:
Will close this. A better solution is to introduce the concept of a middleware where one of the middlewares will be the debugger.

Issue #91, opened 2017-09-27
Passing metadata  through proxy? 


endpress:
I am using grpcwebproxy. My backend server is a node js grpc server. I want pass metadata through   grpc-web-client.

```node js
  const testRequest = new TestRequest();
  testRequest.setUserId('testuserid');

  grpc.unary(MeetService.testHistory, {
    request: testRequest,
    host: host,
    debug: true,
    metadata: {test_token: 'xxxxxxxxxxx'},
    onEnd: res => {
      const { status, statusMessage, headers, message, trailers } = res;
      if (status === Code.OK && message) {
        // console.log('meetHistory.onEnd.message', message.toObject());
      }
      // console.log('meetHistory.onEnd.trailers', trailers);
    }
  });
```

But my node server cannot get metadata. It only can get a user-agent metadata.
```
Metadata { _internal_repr: { 'user-agent': [ 'grpc-go/1.7.0-dev' ] } }
ServerUnaryCall {
  domain: null,
  _events: { error: [Function] },
  _eventsCount: 1,
  _maxListeners: undefined,
  call: Call {},
  cancelled: false,
  metadata: Metadata { _internal_repr: { 'user-agent': [Array] } },
  request: 
   { wrappers_: null,
     messageId_: undefined,
     arrayIndexOffset_: -1,
     array: [ 'testuserid' ],
     pivot_: 1.7976931348623157e+308,
     convertedFloatingPointFields_: {} } }
```

endpress:
My grpc-web-client version is  "grpc-web-client": "^0.3.1",

ypoluektovich:
I'm seeing the same problem. The go proxy thing seems to be swallowing metadata/headers.

suusan2go:
I have the same problem. My backend server is a SpringBoot server(with grpc-java using Kotlin.)

mwitkow:
Indeed. The previous code of grpc-proxy, that the current grpc-webproxy relies on, relied on the implicit forwarding of metadata inside the `context.Context` between gRPC-Go calls. This behaviour has changed in gRPC-Go 1.5. 

The fix requires a change to upstream grpc-go and is discussed in 
https://github.com/mwitkow/grpc-proxy/issues/19

MarcusLongmuir:
[...]

PR #101, opened 2017-11-21
Passing through context/metadata in grpcwebproxy

### LINKED ISSUE #84: Passing token based auth through proxy?
Hello, I'm working with @pjm17971 and this is a follow up to #79.  Thanks for grpc-web -- I think it's fantastic idea and we'd really love to use it in our project.

We're trying to use the grpc-web proxy to talk to an gRPC server that is fronted by [Google Cloud Endpoints](https://cloud.google.com/endpoints/)  -- I've attached a diagram.  We are using JWT tokens for authentication.  It seems like the token isn't making it through the grpcwebproxy.  I've looked at the code and I can't see why that isn't getting passed through.

Do you think the JWT token should be making it through to the backend server?  Is there someplace I'd need to add something to help with that?  I'm totally fine adding some code but I'm not sure I fully get how the proxy works and so I'm not sure if that would be an adjustment to the grpcwebproxy code or perhaps to the grpc-proxy library.  Any pointer you could give would be greatly appreciated! 

esp

MarcusLongmuir:
Using fix in [grpcproxy here](https://github.com/mwitkow/grpc-proxy/pull/20)

Fixes #84 and #91

### DIFF
--- a/go/grpcwebproxy/main.go
+++ b/go/grpcwebproxy/main.go
@@ -23,6 +23,7 @@ import (
 	"golang.org/x/net/context"
 	_ "golang.org/x/net/trace" // register in DefaultServerMux
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


MarcusLongmuir:
Closing in favour of #102

PR #115, opened 2017-12-23
nodeHttp: explicitly set Content-Length

ishitatsuyuki:
This is to avoid unnecessarily making the requests chunked, which are not needed as we don't support client-side streaming yet.

Ref: GAE's frontend is too retarded to handle chunked encoding. https://issuetracker.google.com/issues/71003088

### DIFF
--- a/ts/src/transports/Transport.ts
+++ b/ts/src/transports/Transport.ts
@@ -7,7 +7,7 @@ import {MethodDefinition} from "../service";
 import {ProtobufMessage} from "../message";
 
 export interface Transport {
-  sendMessage(msgBytes: ArrayBufferView): void
+  sendMessage(msgBytes: Uint8Array): void
   finishSend(): void
   cancel(): void
   start(metadata: Metadata): void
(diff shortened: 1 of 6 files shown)


ishitatsuyuki:
ArrayBufferView doesn't make sense and I'll be changing the typing to UInt8Array tomorrow to resolve compile issues.

ishitatsuyuki:
CI green. Please review.

easyCZ:
@ishitatsuyuki Thanks for your PR! Would it be possible to add unit tests to ensure we don't break functionality going forward?

ishitatsuyuki:
Hmm, I'm not sure if unit tests are deemed necessary. If absolutely needed, it would be an assertion on `Content-Length` in Go handler.

easyCZ:
@ishitatsuyuki Could you please rebase the PR?

ishitatsuyuki:
This can be no longer done reasonably due to the interface change; closing.

public:
This makes using the Node transport quite tricky from Python as most WSGI servers have pretty poor chunked encoding support. Would you be interested in a new PR that achieves the same effect as here?

jonnyreeves:
Yes please, although perhaps it could be made configurable when the
transport instance is created?

PR #117, opened 2017-12-27
Fix Grpc-Message decoding

ishitatsuyuki:


ishitatsuyuki:
messages are turned into `string | undefined` due to the old code implicitly violating the contract. Making them null is a better practice though, but as it's backward incompatible some discussion is required.

ishitatsuyuki:
CI green.

ishitatsuyuki:
Review ping.

ts/src/grpc.ts
```diff
@@ -231,7 +235,7 @@ export namespace grpc {
           const gRPCMessage = headers.get("grpc-message") || [];
           props.debug && debug("onHeaders.gRPCMessage", gRPCMessage);
           if (code !== Code.OK) {
-            rawOnError(code, gRPCMessage[0]);
+            rawOnError(code, gRPCMessage.length > 0 ? decodeURIComponent(gRPCMessage[0]) : undefined);
```
easyCZ: `decodeURIComponent` can throw `URIError`, we should attempt to handle it. Could we also use the `""` (empty string) as the value rather than undefined?

ishitatsuyuki:
I have reorganized the code to satisfy the requirements. I tried to minimize code duplication, but basically we have to return early if the decode errors and that part cannot be put inside a function.

ishitatsuyuki:
Oh, \*\*\*\*, the tests are full of `assert.equal` to undefined.

ishitatsuyuki:
This is basically done, but I have no idea why Edge tests are failing. Can you run a retry?

ishitatsuyuki:
The Edge tests seem to be flaky, and I'm probably out of luck. Can you retry again?

easyCZ:
@ishitatsuyuki Sorry for the delay on this. Would you mind rebasing the PR?

ishitatsuyuki:
Sure, I'll do it tomorrow.

ishitatsuyuki:
@easyCZ I think this is ready for review.

test/ts/src/client.spec.ts
```diff
@@ -351,6 +351,39 @@ describe(`client`, () => {
         });
       });
 
+      headerTrailerCombos((withHeaders, withTrailers) => {
+        it("should report status code in Unicode for error with headers + trailers" + name, (done) => {
```
easyCZ: The `name` variable is not defined.

ishitatsuyuki:
All green, please review again.

ts/src/client.ts
```diff
@@ -101,7 +101,14 @@ class GrpcClient<TRequest extends ProtobufMessage, TResponse extends ProtobufMes
       const gRPCMessage = headers.get("grpc-message") || [];
       this.props.debug && debug("onHeaders.gRPCMessage", gRPCMessage);
       if (code !== Code.OK) {
-        this.rawOnError(code, gRPCMessage[0]);
+        let statusMessage;
+        try {
+          statusMessage = this.decodeGRPCStatus(gRPCMessage[0]);
+        } catch (err) {
+          this.rawOnError(Code.Internal, "Cannot decode gRPC-Message header");
```
jonnyreeves: Is this the correct error code? Should it be unknown as it is an error on the transport library rather than the end users code?

ishitatsuyuki: It's unspecified; maybe it should be Unknown but anyway that only happens when a corrupted/broken message is received. Any reference?

jonnyreeves: Only from the discussion in https://github.com/improbable-eng/grpc-web/pull/150

ishitatsuyuki: Aligned the behavior with grpc-go.

PR #120, opened 2018-01-02
Replace UTF-8 decoding with a lightweight library

### LINKED ISSUE #118: Do not use text-encoding for just parsing UTF-8
`text-encoding` became the largest module in my app bundle, 517KB (184KB gzipped). For solely decoding UTF-8, there are some hacks which doesn't bring in a huge dependency.

Reference: https://stackoverflow.com/questions/13356493/decode-utf-8-with-javascript

ishitatsuyuki:
Close #118.

The library used is maintained by protobufjs which is likely to be continuously maintained from now on, and well tested and typed.

### DIFF
--- a/ts/src/ChunkParser.ts
+++ b/ts/src/ChunkParser.ts
@@ -1,9 +1,23 @@
 import {Metadata} from "./grpc";
-import * as TextEncoding from "text-encoding";
 
 const HEADER_SIZE = 5;
 
-const global = Function('return this')();
+const isAllowedControlChars = (char: number) => char === 0x9 || char === 0xa || char === 0xd;
+
+function isValidHeaderAscii(val: number): boolean {
+  return isAllowedControlChars(val) || (val >= 0x20 && val <= 0x7e);
+}
+
+export function decodeASCII(input: Uint8Array): string {
+  // With ES2015, TypedArray.prototype.every can be used
+  for (let i = 0; i !== input.length; ++i) {
+    if (!isValidHeaderAscii(input[i])) {
+      throw new Error("Header is not valid (printable) ASCII");
+    }
+  }
+  // With ES2017, the array conversion can be omitted with iterables
+  return String.fromCharCode(...Array.prototype.slice.call(input));
+}
 
 function isTrailerHeader(headerView: DataView) {
   // This is encoded in the MSB of the grpc header's first byte.
@@ -11,8 +25,7 @@ function isTrailerHeader(headerView: DataView) {
 }
 
 function parseTrailerData(msgData: Uint8Array): Metadata {
-  const decoder = global.TextDecoder !== undefined ? global.TextDecoder : TextEncoding.TextDecoder;
-  return new Metadata(new decoder("utf-8").decode(msgData))
+  return new Metadata(decodeASCII(msgData))
 }
 
 function readLengthFromHeader(headerView: DataView) {
@@ -25,7 +38,7 @@ function hasEnoughBytes(buffer: Uint8Array, position: number, byteCount: number)
 
 export function sliceUint8Array(buffer: Uint8Array, from: number, to?: number) {
   if (buffer.slice) {
-    return buffer.slice(from ,to);
+    return buffer.slice(from, to);
   }
 
   let end = buffer.length;
@@ -36,7 +49,7 @@ export function sliceUint8Array(buffer: Uint8Array, from: number, to?: number) {
   const num = end - from;
   const array = new Uint8Array(num);
   let arrayIndex = 0;
-  for(let i = from; i < end; i++) {
+  for (let i = from; i < end; i++) {
     array[arrayIndex++] = buffer[i];
   }
   return array;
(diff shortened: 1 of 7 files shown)


easyCZ:
Looks good. Will run some tests to check we haven't broken anything. We're also planning to swap `google-protobuf` for `protobufjs` going forward but haven't found the time yet

ishitatsuyuki:
Seems I got the monorepo dependencies incorrect. I will fix it.

easyCZ:
@ishitatsuyuki Could we also please add a test case for decoding UTF-8 to prevent a regression?

ishitatsuyuki:
Actually, I'm pretty sure that non-printable ASCII characters are not allowed in headers/metadata. Thus, I'm not really sure if the UTF-8 decoding is even needed here.

ishitatsuyuki:
I have reimplemented it with an assertion of ASCII. I hope it works.

ishitatsuyuki:
Something went terribly for IE11. What do you think is the reason?

ishitatsuyuki:
Friendly ping for review, @easyCZ

ishitatsuyuki:
First, I replaced it with an ASCII decoder, which is already tested as a part of trailer parsing.

Second, you don't seem to have any unit tests, and no facility for testing non-exported functions.

ts/src/ChunkParser.ts
```diff
@@ -1,18 +1,25 @@
 import {Metadata} from "./grpc";
-import * as TextEncoding from "text-encoding";
 
 const HEADER_SIZE = 5;
 
-const global = Function('return this')();
+function decodeASCII(input: Uint8Array): string {
```
easyCZ: We can pull this into a module, something like `metadata.ts`. With that, we can simplify the method name to be just `export function decode(input: Uint8Array) {...}`. We gain the ability to test the implementation.

With regards to correctness, the [spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md) outlines that the possible values can be the following:
* Custom-Metadata → Binary-Header / ASCII-Header
* Binary-Header → {Header-Name "-bin" } {base64 encoded value}
* ASCII-Header → Header-Name ASCII-Value
* Header-Name → 1*( %x30-39 / %x61-7A / "_" / "-" / ".") ; 0-9 a-z _ - .
* ASCII-Value → 1*( %x20-%x7E ) ; space and printable ASCII

You're absolutely correct in suggesting that UTF-8 decoding is not necessary as it is not allowed by the spec.

In your implementation, we are missing the handling of all `printable` characters, that is anything between `0x9` through to `0xD`.
We can do something like this:
```javascript
// tab, newline, verticaltab, formfeed, carriage return
const isSpace = (char: number) => char >= 0x9 && char <= 0xD

function isValidGrpcAsciiValue(val: number): boolean {
    return isSpace(val) || (val >= 0x20 && val <= 0x7e)
}
```

Furthermore, I propose that instead of throwing an error (which we'll need to catch), we drop the invalid characters and log a warning for the engineers. Something like this:
```javascript
export function decode(input: Uint8Array): string {
  return input
    .filter(val => {
      const isValid = isValidGrpcAsciiValue(val);
      if (!isValid) {
        console.warn(`Encountered invalid character in metadata. Attempted to decode ${input} but ${val} is not a valid grpc metadata character. Dropping the character.`)
      }
      return isValid;
    })
    .map(val => ...)

}
```

Let me know what you think

ishitatsuyuki: I think it's a good idea, but however it's not really easy to drop characters in ES5 (almost no extension methods on typed arrays).

As a result, I think bailing out with an error is the simplest solution here.

ishitatsuyuki: Uh, and also, are characters other than CRLF allowed in HTTP 1.1 headers? I initially guessed it's not.
(The HTTP2 spec uses binary framing so none of those characters are used and thus not documented.)

easyCZ: Looks like you're right, the HTTP 1.1 spec says 
```
3.2.3.  Whitespace

   This specification uses three rules to denote the use of linear
   whitespace: OWS (optional whitespace), RWS (required whitespace), and
   BWS ("bad" whitespace).

   The OWS rule is used where zero or more linear whitespace octets
   might appear.  For protocol elements where optional whitespace is
   preferred to improve readability, a sender SHOULD generate the
   optional whitespace as a single SP; otherwise, a sender SHOULD NOT
   generate optional whitespace except as needed to white out invalid or
   unwanted protocol elements during in-place message filtering.

   The RWS rule is used when at least one linear whitespace octet is
   required to separate field tokens.  A sender SHOULD generate RWS as a
   single SP.

   The BWS rule is used where the grammar allows optional whitespace
   only for historical reasons.  A sender MUST NOT generate BWS in
   messages.  A recipient MUST parse for such bad whitespace and remove
   it before interpreting the protocol element.

     OWS            = *( SP / HTAB )
                    ; optional whitespace
     RWS            = 1*( SP / HTAB )
                    ; required whitespace
     BWS            = OWS
                    ; "bad" whitespace
```
 so only `SP` or `HTAB` are allowed

ishitatsuyuki:
Added tests, I think this one is OK.

PR #132, opened 2018-01-23
Refactored to expose client()

MarcusLongmuir:
This PR is a significant refactor to expose a "client" from a new `client()` function to enable a websocket transport that will bring bi-directional gRPC support.

In doing so I've:
* Written API docs for the functions
* Updated READMEs
* Refactored the testing to enable overriding transports (to be used for testing future websocket transport)
* Refactored the index file (grpc namespace)
* Explained transports and other concepts
* Extended the browser testing


test/ts/src/testCombinations.ts
```diff
@@ -0,0 +1,65 @@
+import {
```
easyCZ: Perhaps a better filename here, something like `testRpcCombinations` to make it more explicit

ts/src/client.ts
```diff
@@ -0,0 +1,311 @@
...
+  onHeaders(callback: (headers: Metadata) => void): void;
+  onMessage(callback: (message: TResponse) => void): void;
+  onEnd(callback: (code: Code, message: string, trailers: Metadata) => void): void;
+}
+
+export function client<TRequest extends ProtobufMessage, TResponse extends ProtobufMessage, M extends MethodDefinition<TRequest, TResponse>>(methodDescriptor: M, props: ClientRpcOptions): Client<TRequest, TResponse> {
+  return new ClientImpl<TRequest, TResponse, M>(methodDescriptor, props);
+}
+
+class ClientImpl<TRequest extends ProtobufMessage, TResponse extends ProtobufMessage, M extends MethodDefinition<TRequest, TResponse>> {
```
easyCZ: [nit] Perhaps `GrpcClient`?

ts/src/client.ts
```diff
@@ -0,0 +1,311 @@
...
+  methodDefinition: M;
+  props: ClientRpcOptions;
+
+  started: boolean = false;
+  sentFirstMessage: boolean = false;
+  completed: boolean = false;
+  closed: boolean = false;
+  finishedSend: boolean = false;
+
+  onHeadersCallbacks: Array<(headers: Metadata) => void> = [];
```
easyCZ: Pulling this into a middleware in the future might make more sense, the client then only has one callback for each life cycle method.

Right now we also don't have a way to remove callbacks that may already have been registered. I think this fine as long as we know it's a limitation

ts/src/client.ts
```diff
@@ -0,0 +1,311 @@
...
+        callback(res);
+      });
+    });
+  }
+
+  onHeaders(callback: (headers: Metadata) => void) {
+    this.onHeadersCallbacks.push(callback);
+  }
+
+  onMessage(callback: (res: TResponse) => void) {
```
easyCZ: Would it make sense to prevent attaching callbacks once a request has started?

ts/src/client.ts
```diff
@@ -0,0 +1,311 @@
...
+
+class ClientImpl<TRequest extends ProtobufMessage, TResponse extends ProtobufMessage, M extends MethodDefinition<TRequest, TResponse>> {
+  methodDefinition: M;
+  props: ClientRpcOptions;
+
+  started: boolean = false;
+  sentFirstMessage: boolean = false;
+  completed: boolean = false;
+  closed: boolean = false;
+  finishedSend: boolean = false;
```
easyCZ: [nit] `finishedSending`?

ts/src/index.ts
```diff
@@ -1,9 +1,39 @@
...
+  export interface ClientRpcOptions extends impClient.ClientRpcOptions{}
+
+  export const invoke = impInvoke.invoke;
+  export interface Request extends impInvoke.Request{}
+  export interface InvokeRpcOptions<TRequest extends ProtobufMessage, TResponse extends ProtobufMessage> extends impInvoke.InvokeRpcOptions<TRequest, TResponse>{}
+
+  export const unary = impUnary.unary;
+  export interface UnaryOutput<TResponse extends ProtobufMessage> extends impUnary.UnaryOutput<TResponse>{}
+  export interface UnaryRpcOptions<TRequest extends ProtobufMessage, TResponse extends ProtobufMessage> extends impUnary.UnaryRpcOptions<TRequest, TResponse>{}
+}
```
easyCZ: 👍

ts/docs/client.md
```diff
@@ -0,0 +1,86 @@
...
+#### `Client`:
+```typescript
+// Open the connection to the server
+start(metadata?: grpc.Metadata): void;
+
+// Send a single instance of the request message type
+send(message: grpc.ProtobufMessage): void;
+
+// Indicate to the server that the client has finished sending
+finishSend(): void;
```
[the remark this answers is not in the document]
MarcusLongmuir: `finishSend()` is used by the client to indicate to the server that it's finished sending - it doesn't actually end the lifecycle of the rpc; in some usages the server only starts sending messages when the client indicates that it has finished sending.

I think naming this `end()` will make this less clear as it might be assumed that this function will just terminate the rpc.

PR #137, opened 2018-02-07
Added experimental Websocket transport

MarcusLongmuir:


go/grpcweb/websocket_wrapper.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+func (w *WebSocketResponseWriter) extractTrailerHeaders() http.Header {
+	trailerHeaders := make(http.Header)
+	for k, vv := range w.headers {
+		// Skip the pre-annoucement of Trailer headers. Don't add them to the response headers.
+		if strings.ToLower(k) == "trailer" {
+			continue
+		}
+		//// Skip existing headers that were already sent.
```
MarcusLongmuir: TODO

go/grpcweb/websocket_wrapper.go
```diff
@@ -0,0 +1,204 @@
...
+
+func (w *WebSocketWrappedReader) Close() error {
+	w.respWriter.FlushTrailers()
+	return w.wsConn.Close()
+}
+
+// First byte of a binary WebSocket frame is used for control flow:
+// 0 = Data
+// 1 = End of client send
+func (w *WebSocketWrappedReader) Read(p []byte) (int, error) {
```
MarcusLongmuir: Here's the interesting bit

go/grpcweb/wrapper.go
```diff
@@ -10,7 +10,10 @@ import (
 	"time"
 
 	"github.com/rs/cors"
+	"github.com/gorilla/websocket"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/grpclog"
+	"net/url"
```
johanbrandhorst: This import should be with the other standard library imports

go/grpcweb/websocket_wrapper.go
```diff
@@ -0,0 +1,204 @@
+package grpcweb
+
+import (
+	"bufio"
+	"bytes"
+	"encoding/binary"
+	"github.com/gorilla/websocket"
```
johanbrandhorst: This import should not be with the standard library imports

go/grpcweb/wrapper.go
```diff
@@ -57,13 +76,29 @@ func WrapServer(server *grpc.Server, options ...Option) *WrappedGrpcServer {
...
 // You can control the CORS behaviour using `With*` options in the WrapServer function.
 func (w *WrappedGrpcServer) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
+	if w.enableWebsockets && w.IsGrpcWebSocketRequest(req) {
+		if w.websocketOriginFunc(req) {
+			if !w.opts.corsForRegisteredEndpointsOnly || w.isRequestForRegisteredEndpoint(req) {
+				w.HandleGrpcWebsocketRequest(resp, req)
+				return
+			}
+		}
+		resp.WriteHeader(400)
```
johanbrandhorst: `StatusBadRequest`? Doesn't `StatusForbidden` (403) make more sense here? 403 instead of 401 because it's Forbidden from this Origin and not a matter of authentication.

go/grpcweb/wrapper.go
```diff
@@ -74,6 +109,56 @@ func (w *WrappedGrpcServer) HandleGrpcWebRequest(resp http.ResponseWriter, req *
...
+
+	headers, err := parseHeaders(string(readBytes))
+	if err != nil {
+		grpclog.Errorf("Unable to parse websocket headers: %v", err)
+		return
+	}
+
+	wrappedReader := NewWebsocketWrappedReader(wsConn, respWriter)
+
+	req.Body = wrappedReader
```
johanbrandhorst: This is mind blowing. Well done!

johanbrandhorst:
What's the status of this? Doesn't look too bad to me but no updates for a week and a half.

easyCZ:
I'm looking to do a review and some testing with this over the weekend. No concrete ETA on this yet.

johanbrandhorst:
Ok cool, I'll probably start integrating it into downstream (my gopherjs library) based on the contents of this branch, it's a much cleverer implementation than my current websocket proxy. Just excited to see this in here :).

abeMedia:
Awesome work done here! Correct me if I'm wrong but for this to be useful the grpc-web typescript packages would need updating to support websockets too, right?

go/grpcweb/wrapper.go
```diff
@@ -41,11 +46,25 @@ func WrapServer(server *grpc.Server, options ...Option) *WrappedGrpcServer {
 		AllowCredentials: true,                                // always allow credentials, otherwise :authorization headers won't work
 		MaxAge:           int(10 * time.Minute / time.Second), // make sure pre-flights don't happen too often (every 5s for Chromium :( )
 	})
+	websocketOriginFunc := opts.websocketOriginFunc
+	if websocketOriginFunc == nil {
+		websocketOriginFunc = func(req *http.Request) bool {
```
easyCZ: Could we pull this into a separate function? Something like `defaultWebsocketOriginFunc`

go/grpcweb/wrapper.go
easyCZ: Agreed, 403 is better.

test/ts/src/client.spec.ts
```diff
@@ -525,6 +528,181 @@ describe(`client`, () => {
         });
       });
     });
+
+    if (!process.env.DISABLE_WEBSOCKET_TESTS) {
+      describe("client-streaming (websockets)", () => {
```
easyCZ: Would it make sense to pull this into a separate file?

johanbrandhorst:
Tested this in my downstream projects, works well :).

ts/src/transports/Transport.ts
```diff
@@ -59,3 +60,7 @@ function detectTransport(): TransportConstructor {
 
   throw new Error("No suitable transport found for gRPC-Web");
```
johanbrandhorst: Any chance we could add the WebsocketTransport to auto-detected transports?

johanbrandhorst:
https://github.com/improbable-eng/grpc-web/pull/137/files#diff-d81ccdf67913238d5ac6fce3cb4ed098R34 this isn't true anymore

go/grpcweb/options.go
```diff
@@ -76,3 +80,22 @@ func WithAllowedRequestHeaders(headers []string) Option {
 		o.allowedRequestHeaders = headers
 	}
 }
+
+// WithWebsockets allows for handling grpc-web requests of websockets - enabling bidirectional requests.
+//
+// The default behaviour is false, i.e. to disallow websockets
+func WithWebsockets(enableWebsockets bool) Option {
```
johanbrandhorst: I'd be keen to just turn this on, it's not a backwards compatibility breaking change, and unless my suggestion to add Websockets to DetectTransport is implemented it won't change anything.

ts/src/transports/Transport.ts
MarcusLongmuir: I'd be keen to separate the implementation of the transport (this PR) from a discussion around how this should be used.

Currently the default transport factory just chooses the transport through a fallback process, but Websockets aren't necessarily the first in that list of transports to fallback through as they have some potentially undesirable tradeoffs.

I'd be happy to look at a PR that proposes a good standard way of using this transport after this lands 👍

johanbrandhorst:
I've discovered a bit of an inconsistency with regards to headers:

When using the MozXHR transport, the headers received look like this (both in the case of unary and server streaming calls):
```
rawOnHeaders {
    headersMap: {
        "content-type": Array [ "application/octet-stream" ]​​
    }​
}
```

When using the websocket transport, the headers received look like this:
```
rawOnHeaders {
​    headersMap: {
        "content-type": Array [ "application/grpc" ]​​
        trailer: Array(3) [ "Grpc-Status", "Grpc-Message", "Grpc-Status-Details-Bin" ]​​
    }
}
```

So the curious thing to me is that we see the `trailer` in `rawOnHeaders`. Not sure we can or should do anything about it, but would be cool to hear your thoughts on this.

johanbrandhorst:
Disregard my previous comment, it's not particularly interesting. I've tested this pretty extensively and would be keen to see it merged sooner rather than later :).

DieMyst:
hello. What is about this PR? No updates for 10 days. Can I use this branch for prototyping for now?

johanbrandhorst:
@MarcusLongmuir has been pushing things recently, so I remain hopeful this will be merged soon :)

MarcusLongmuir:
@johanbrandhorst + @DieMyst: Apologies for delay. Testing this has presented some issues - Browserstack's testing tunnel, websockets and self-signed certs have taken some effort to root cause and overcome. Hoping to get this merged ASAP.

johanbrandhorst:
Thanks for this, super excited to expose it in my downstream package :)

DieMyst:
Good news! Big thanks @MarcusLongmuir

PR #140, opened 2018-02-08
Adapt for AppEngine

maxhille:
I want to use your GRPC-WEB implementation on AppEngine. Since the underlying (internal, proprietary) HTTP context there does not implement http.Flusher and http.CloseNotifier, I had to modify the code by making there interfaces optional. The behaviour now matches the docs in http-std for these interfaces.

Not being able to look at AppEngines actual code, I suggest the limitations come from AppEngine not supporting any kind of server side streaming. As such streams will just pile up and sent in one go. I do not think of this as a serious problem.

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 [![Master Build](https://travis-ci.org/improbable-eng/grpc-web.svg)](https://travis-ci.org/improbable-eng/grpc-web)
 ![BrowserStack Status](https://www.browserstack.com/automate/badge.svg?badge_key=L0k3QjhiNnByd3hWVVhtS0FxTmNrZERwbDBqR053OFJKV01veUpkL1FqOD0tLXZyS0d2WC9TaGEzeTBjbXZ6L1JNa2c9PQ==--b460187586f63fc2a48f557a515f9900f5639d10)
 [![NPM](https://img.shields.io/npm/v/grpc-web-client.svg)](https://www.npmjs.com/package/grpc-web-client)
-[![GoDoc](http://img.shields.io/badge/GoDoc-Reference-blue.svg)](https://godoc.org/github.com/improbable-eng/grpc-web/go/grpcweb) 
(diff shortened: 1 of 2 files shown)


maxekman:
I would benefit from this change as I’m just experimenting with using grpc-web for a client’s project.

I think the lack of streaming is an acceptable limitation when using on App Engine, but it would be good to make a note about that in the limitations section in the README.

maxhille:
Just as a heads-up you will likely run into another issue when using this on AppEngine: The context in your handler implementation will wrap the AppEngine context somehow internally. I did not find a way to get it out and even if I would have succeeded that would have likely been a hacky use of internal API. Instead I chose to inject the context with my own key[1]

[1] https://github.com/lambdasoup/finmgr/blob/master/aegrpc/aegrpc.go

maxekman:
Ah, thanks. That would indeed be tricky to find myself!

README.md
```diff
@@ -166,6 +166,11 @@ XHR keeps the entire server response in memory. This means that a long-lived or
 
 You can read more about how grpc-web-client determines and uses transports [here](ts/docs/transport.md).
 
+### AppEngine
+
+AppEngine does not support server side streaming. The streaming response API will work but end up as one big response on the client side. As usual with AppEngine, the server code should try to finish the response as quick as possible to stay within deadlines.
```
maxekman: Perfect!

maxhille:
I tried to fix the tests now but did not come very far:

* Locally, I can run with Firefox/Ubuntu and Chrome/Ubuntu
* Firefox here is all green, Chrome has a deterministic failure which does not show up on Travis
* The Travis failures (IE and safari) will not be repoducible for me locally

So I guess I am kinda stuck

ishitatsuyuki:
https://github.com/k2wanko/gaegrpc/ is another working wrapper for Standard Environment.

maxhille:
I took a look at the wrapper provided by @ishitatsuyuki. I will close this PR because the mentioned solution is cleaner than this one I think.

PR #142, opened 2018-02-17
Fix ts errors by updating grpc web client version

dangduc:
Due to a major refactoring of the typescript client code between 0.4.0 and 0.5.0, the example project typescript code was *updated* but there needs to be a corresponding module version bump in order to clear up some syntax errors. This commit fixes `npm start` for the example.

### DIFF
--- a/example/package.json
+++ b/example/package.json
@@ -15,7 +15,7 @@
   "dependencies": {
     "@types/google-protobuf": "^3.2.5",
     "google-protobuf": "^3.2.0",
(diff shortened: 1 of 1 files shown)


johanbrandhorst:
This is a duplicate of https://github.com/improbable-eng/grpc-web/pull/138

easyCZ:
This is indded a duplicate. Closing as #138 also updates the package-lock which would be required for this PR.

PR #147, opened 2018-03-04
Fix error in debug log message

johanbrandhorst:
In my testing of https://github.com/improbable-eng/grpc-web/pull/137/files I ran into an instance of this where `res` did not have a function called `toObject`. This fix removes that assumption, though my understanding of TypeScript is that it shouldn't be possible in the first place? I experienced this on Firefox 59.


johanbrandhorst:
Nevermind, I totally know why this happened.

Issue #145, opened 2018-02-27
proxy: stream terminated by RST_STREAM


jonahbron:
I'm trying to get a basic example of grpc-web working in my application.  Using the HelloWorld service example from the grpc-node library, and the example code from grpc-web/example, I've put together what I think should work.  The proxy connects to my server...

```
time="2018-02-27T18:22:42Z" level=info msg="dialing to target with scheme: \"\"" system=system
time="2018-02-27T18:22:42Z" level=info msg="listening for http_tls on: [::]:8443"
time="2018-02-27T18:22:42Z" level=info msg="ccResolverWrapper: sending new addresses to cc: [{stern-server:9000 0  <nil>}]" system=system
time="2018-02-27T18:22:42Z" level=info msg="ClientConn switching balancer to \"pick_first\"" system=system
time="2018-02-27T18:22:42Z" level=info msg="listening for http on: [::]:8080"
time="2018-02-27T18:22:42Z" level=info msg="pickfirstBalancer: HandleSubConnStateChange: 0xc42016e370, CONNECTING" system=system
time="2018-02-27T18:22:42Z" level=info msg="pickfirstBalancer: HandleSubConnStateChange: 0xc42016e370, READY" system=system
```

When I use the client to make the call, it creates a request like this (copied as cURL from Chrome network log):

    curl 'https://localhost:8443/helloworld.Greeter/SayHello' -H 'origin: https://localhost:4200' -H 'accept-encoding: gzip, deflate, br' -H 'accept-language: en-US,en;q=0.9' -H 'user-agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/64.0.3282.167 Safari/537.36' -H 'content-type: application/grpc-web+proto' -H 'accept: */*' -H 'x-grpc-web: 1' -H 'referer: https://localhost:4200/operations' -H 'authority: localhost:8443' --data-binary $'\x00\x00\x00\x00\x04\n\x02yo' --compressed --insecure

But this is the error message it gives back:

    stream terminated by RST_STREAM with error code: INTERNAL_ERROR

That same error shows up in the log output from the proxy itself.

I know my server should work because I created a grpc-node client and it was able to connect just fine.

What could this error indicate?  Where should I start looking?  I've run out of ideas at this point.

jonahbron:
[...]

PR #148, opened 2018-03-16
145 fix bug proxy: stream terminated by RST_STREAM

MrWinstead:
* Apply patch suggested by @hpidcock at
https://github.com/improbable-eng/grpc-web/issues/145#issuecomment-371716140

### DIFF
--- a/go/grpcwebproxy/main.go
+++ b/go/grpcwebproxy/main.go
@@ -104,7 +104,9 @@ func buildGrpcProxyServer(logger *logrus.Entry) *grpc.Server {
 	director := func(ctx context.Context, fullMethodName string) (context.Context, *grpc.ClientConn, error) {
 		md, _ := metadata.FromIncomingContext(ctx)
 		outCtx, _ := context.WithCancel(ctx)
-		outCtx = metadata.NewOutgoingContext(outCtx, md.Copy())
+    mdCopy := md.Copy()
+    delete(mdCopy, "user-agent")
+		outCtx = metadata.NewOutgoingContext(outCtx, mdCopy)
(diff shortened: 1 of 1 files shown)


MrWinstead:
I'm currently using this as part of an internal project and the fix allows a javascript gRPC client to communicate from a ReactJS Redux action to a python 3 gRPC server.

go/grpcwebproxy/main.go
```diff
@@ -104,7 +104,9 @@ func buildGrpcProxyServer(logger *logrus.Entry) *grpc.Server {
 	director := func(ctx context.Context, fullMethodName string) (context.Context, *grpc.ClientConn, error) {
 		md, _ := metadata.FromIncomingContext(ctx)
 		outCtx, _ := context.WithCancel(ctx)
-		outCtx = metadata.NewOutgoingContext(outCtx, md.Copy())
+    mdCopy := md.Copy()
```
jonahbron: Some weird indentation going on here.

jonahbron:
This is brilliant.  The error that CI is picking up is:

          ✖ should handle aborting a streaming response mid-stream with propagation of the disconnection to the server

But I seriously doubt this patch has caused that regression, I suspect that error was already there.  Can this be merged in soon so I can run my proxy on the latest code?

MrWinstead:
Thank you! I've updated the indentation. At first I just made the quick edit in vi, but now I finally opened the poor thing in Goland to correct things

jonahbron:
@MrWinstead Still looks off in the Github diff.  I think the surrounding code uses actual tab characters, while yours uses spaces.

PR #150, opened 2018-03-17
Make gRPC status codes consistent with gRPC-Go

### LINKED ISSUE #149: Status code used when network error occurs is inconsistent with Go gRPC Client
Hi,

As part of my downstream project [`gRPC-Web for Go`](https://github.com/johanbrandhorst/protobuf/)'s migration to the new websocket transport, I developed a testing framework that allows me to run the same tests against the gRPC-Web client and the Go gRPC client. Please see the source ([gRPC-Web](https://github.com/johanbrandhorst/protobuf/blob/21e69814b025a0e57181cddb093908f3d934c8c1/test/client/main.go#L337), [gRPC-Go](https://github.com/johanbrandhorst/protobuf/blob/6e3e35cbb49a72e95c4815ce87d1256f5304e630/test/webdriver_test.go#L180)) for more details.

Through this, I've noticed there's an inconsistency when it comes to returned status codes from the client when the network connection is dropped. gRPC-Go returns [`codes.Unknown`](https://github.com/grpc/grpc-go/blob/fa28bef9392c6c3e28e75389d8be8a6797561f57/transport/http_util.go#L268), and gRPC-Web returns [`codes.Internal`](https://github.com/improbable-eng/grpc-web/blob/eb398c2f9233171176ceebb2353148224a0e58a0/ts/src/client.ts#L165).

Ideally the gRPC-Web client should return `codes.Unknown`, in my opinion.

johanbrandhorst:
This changes the two status errors that can be returned during network errors to `codes.Unknown`, which is consistent with the gRPC-Go client implementation.

Fixes #149

### DIFF
--- a/test/ts/src/client.spec.ts
+++ b/test/ts/src/client.spec.ts
@@ -378,7 +378,7 @@ describe(`client`, () => {
             DEBUG && debug("status", status, "statusMessage", statusMessage, "trailers", trailers);
             // Some browsers return empty Headers for failed requests
             assert.strictEqual(statusMessage, "Response closed without headers");
-            assert.strictEqual(status, grpc.Code.Internal);
+            assert.strictEqual(status, grpc.Code.Unknown);
             assert.ok(!didGetOnMessage);
             done();
           });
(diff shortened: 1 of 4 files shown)


jonny-improbable:
Thanks @johanbrandhorst, LGTM. 👍

PR #160, opened 2018-04-02
Lint Typescript sources

jonny-improbable:
- Include typescript in linting checks (previously only linted go sources)
- Modify Travis config so we don't retry on linting failures (only on test runs)
- Fix pre-existing Typescript linting errors


test/ts/src/client.spec.ts
```diff
@@ -4,8 +4,8 @@ import {grpc} from "../../../ts/src/index";
 import {debug} from "../../../ts/src/debug";
 import {assert} from "chai";
 // Generated Test Classes
-import {Empty,} from "google-protobuf/google/protobuf/empty_pb";
-import {PingRequest, PingResponse,} from "../_proto/improbable/grpcweb/test/test_pb";
+import {Empty, } from "google-protobuf/google/protobuf/empty_pb";
```
easyCZ: This linting looks odd. Will cut a follow up ticket to clean it up.

PR #166, opened 2018-04-11
Change visibility of new GRPC WebSocket functions.

jonny-improbable:
@MarcusLongmuir Looking at #137 it appeared a lot of new functions were exported into the public API; this PR removes them - LMK if you exported them for a reason (in which case we should document them!) :)

### DIFF
--- a/go/grpcweb/websocket_wrapper.go
+++ b/go/grpcweb/websocket_wrapper.go
@@ -14,11 +14,7 @@ import (
 	"golang.org/x/net/http2"
 )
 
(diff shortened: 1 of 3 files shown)


go/grpcweb/websocket_wrapper.go
```diff
@@ -14,20 +14,16 @@ import (
 	"golang.org/x/net/http2"
 )
 
-type WebSocketWrapper struct {
```
jonny-improbable: This appeared to be unused.

PR #173, opened 2018-04-19
Added abort signal to fetch transport

Runar1:
Added abort signal to fetch transport so that we can cancel requests before any data is sent/received.
Chrome support for AbortController is coming in version 66.


ts/src/transports/fetch.ts
```diff
@@ -44,12 +45,13 @@ class Fetch implements Transport {
       });
   }
 
-  send(msgBytes: ArrayBufferView) {
+  send(msgBytes: Uint8Array) {
```
MarcusLongmuir: This makes sense as `fetch` now only accepts `Uint8Array` and not `ArrayBufferView`. Can you replace `ArrayBufferView` with `Uint8Array` across the codebase in this change please?

PR #176, opened 2018-04-21
Upgrade go dependencies

easyCZ:
Ran `dep ensure -update`


easyCZ:
Good point, will close for now

PR #178, opened 2018-04-21
Re-add "Fix Grpc-Message decoding (#117)"

easyCZ:
This re-adds commit 9df9c0a453673a3ceed1b91a794d731834bf54c7.

@ishitatsuyuki Had to revert the change from #117 due to a breaking build. Will use this PR going forward.


MarcusLongmuir:
Looks like `Uncaught AssertionError: expected no message: expected '' to equal undefined thrown` is being thrown on almost all test runs.

ts/src/client.ts
```diff
@@ -179,7 +181,20 @@ class GrpcClient<TRequest extends ProtobufMessage, TResponse extends ProtobufMes
...
+
+  decodeGRPCStatus(src: string | undefined): string {
+    if (src) {
+      try {
+        return decodeURIComponent(src)
+      } catch (err) {
+        return src
+      }
+    } else {
+      return ""
```
MarcusLongmuir: Is there a reason to return empty string here instead of `null` or `undefined` (undefined was the previous behaviour)? 

Appreciate that this is likely just opinion, but this is a subtle API change and the reason this PR is currently failing tests.

PR #184, opened 2018-04-29
Prevent fetch.pump recursively returning promises

### LINKED ISSUE #183: Memory leak inside Fetch pump method
[Fetch.prototype.pump](https://github.com/improbable-eng/grpc-web/blob/master/ts/src/transports/fetch.ts#L25) is recursively calling itself internally inside of the [this.reader.read()](https://github.com/improbable-eng/grpc-web/blob/master/ts/src/transports/fetch.ts#L43) promise for streamed requests.

Due to the pump method returning [this.reader.read()](https://github.com/improbable-eng/grpc-web/blob/master/ts/src/transports/fetch.ts#L32) I think it may be building up an infinite promise chain.

Below are 2 images of memory allocation timelines, one with the `return` statement removed. (I also had to comment out `_this.options.onChunk(result.value);` as it looks like there's a leak within that too.)

With promise return:
with promise return

without promise return:
without promise return

nathanb21:
Fixes #183

### DIFF
--- a/ts/src/transports/fetch.ts
+++ b/ts/src/transports/fetch.ts
@@ -22,14 +22,15 @@ class Fetch implements Transport {
     this.options = transportOptions;
   }
 
-  pump(readerArg: ReadableStreamReader, res: Response): Promise<void | Response> {
+  pump(readerArg: ReadableStreamReader, res: Response) {
     this.reader = readerArg;
     if (this.cancelled) {
       // If the request was cancelled before the first pump then cancel it here
       this.options.debug && debug("Fetch.pump.cancel at first pump");
-      return this.reader.cancel();
+      this.reader.cancel();
+      return;
     }
-    return this.reader.read()
+    this.reader.read()
       .then((result: { done: boolean, value: Uint8Array }) => {
         if (result.done) {
           detach(() => {
@@ -40,7 +41,8 @@ class Fetch implements Transport {
         detach(() => {
           this.options.onChunk(result.value);
         });
-        return this.pump(this.reader, res);
+        this.pump(this.reader, res);
+        return;
       });
   }
 
@@ -56,7 +58,8 @@ class Fetch implements Transport {
         this.options.onHeaders(new Metadata(res.headers as any), res.status);
       });
       if (res.body) {
-        return this.pump(res.body.getReader(), res)
+        this.pump(res.body.getReader(), res)
+        return;
       }
       return res;
     }).catch(err => {


jonnyreeves:
Thanks for this patch, please could you include a screenshot of the chrome memory profiler which demonstrates the new characteristics with this change.

ts/src/transports/fetch.ts
```diff
@@ -22,14 +22,15 @@ class Fetch implements Transport {
     this.options = transportOptions;
   }
 
-  pump(readerArg: ReadableStreamReader, res: Response): Promise<void | Response> {
+  pump(readerArg: ReadableStreamReader, res: Response) {
     this.reader = readerArg;
```
jonnyreeves: Could you explicitaly type the return value, please?

ts/src/transports/fetch.ts
```diff
@@ -40,7 +41,8 @@ class Fetch implements Transport {
         detach(() => {
           this.options.onChunk(result.value);
         });
-        return this.pump(this.reader, res);
+        this.pump(this.reader, res);
+        return;
```
jonnyreeves: I think this return statement is redundant and can be removed

ts/src/transports/fetch.ts
nathanb21: Pump no longer returns anything. The method now matches the [send](https://github.com/improbable-eng/grpc-web/blob/master/ts/src/transports/fetch.ts#L47) method within the class.

The previously returned promise was never actually used anywhere. The pump method was only called recursively, or from within the `send` method; and send has no return value.

jonnyreeves: > All methods within this class that aren't returning anything are missing the void return type

🤔 Apologies I assumed we had explicit typing. 

Given that we do not this boils down to personal preference so feel free to disregard my comment.

ts/src/transports/fetch.ts
nathanb21: You're right, have removed.

nathanb21:
@jonnyreeves added screenshots taken from issue 183. Thanks for your feedback 👍

ts/src/transports/fetch.ts
nathanb21: Removing this caused CI to fail - 'Not all code paths return a value'

jonnyreeves: Oh dear that's 2/2 for bad suggestions from me 😂 Apologies for any time wasted and thanks again for the patch

jonnyreeves:
Good to merge. Will aim to action this and cut a release to npm either today or tomorrow, schedule permitting. Thanks again

PR #188, opened 2018-05-02
Updated example to 0.6.2 and updated dependencies

MarcusLongmuir:


example/ts/webpack.config.js
```diff
@@ -3,6 +3,7 @@ const path = require('path');
 
 module.exports = {
   entry: "./src/index.ts",
+  mode: "development",
```
thomasetter: Thanks for updating the example!

I tried this on my machine and this line caused this error message:
```
[1] Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
[1]  - configuration has an unknown property 'mode'. These properties are valid:
[1]    object { amd?, bail?, cache?, context?, dependencies?, devServer?, devtool?, entry, externals?, loader?, module?, name?, node?, output?, performance?, plugins?, profile?, recordsInputPath?, recordsOutputPath?, recordsPath?, resolve?, resolveLoader?, stats?, target?, watch?, watchOptions? }
```
Without this line, it runs fine.
Should we keep this as it is now?

MarcusLongmuir: Did you `npm install` first? I've just tried this again on a fresh install and it works fine. Please raise an issue / PR if you reproduce this.

PR #189, opened 2018-05-04
Fixed bug where we got unhandled promise rejections in fetch pump

Runar1:
After the latest update I got unhandled promise rejections in the fetch pump when the server closed the connection for some reason.


ts/src/transports/fetch.ts
```diff
@@ -44,6 +44,16 @@ class Fetch implements Transport {
         });
         this.pump(this.reader, res);
         return;
+      })
+      .catch(err => {
+        if (this.cancelled) {
+          this.options.debug && debug("Fetch.catch - request cancelled");
```
easyCZ: Would we need to invoke `this.reader.cancel()` here as well?

PatrickLehnerXI:
any news on this?

since we can't currently handle all errors due to this, we can't really use grpc-web-client at the moment :( -- but we would really like to

laSinteZ:
I agree. Is there any news? This is pretty important feature for us as well :(

laSinteZ:
@jonny-improbable sorry, but will you release the fixed version to npm? ^^

PR #201, opened 2018-06-07
Keep original headers

crazed:
I was giving this project a shot and have to say it's pretty nice doing JS -> grpc. 

I don't know if this is the right approach here but there are custom headers that I would like to expose to grpc metadata so that in an interceptor I can access these values (for example JWT).

This allows headers to propagate all the way down rather than being overwritten. Let me know if there's a better way to handle this.

### DIFF
--- a/go/grpcweb/wrapper.go
+++ b/go/grpcweb/wrapper.go
@@ -148,12 +148,16 @@ func (w *WrappedGrpcServer) handleWebSocket(wsConn *websocket.Conn, req *http.Re
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


crazed:
@johanbrandhorst no problem, rebased

johanbrandhorst:
Yes, could we add a test for this? I assume there should be one for the other transports already so might be able to copy one and just make it use the WS transport.

crazed:
Yeah I can add a test here, might take a day or two depending on work schedule.

PR #204, opened 2018-06-08
Removed Node transport from automatic detection

jonahbron:
This is an alternative to https://github.com/improbable-eng/grpc-web/pull/203/files, and arguably a better solution.  The downside is that this is a breaking change for any Node users, which means merging this will require a major version increment.  This change simply requires that the developer import the Node transport if they want it, instead of automatically detecting it.  This ensures the Node transport can be omitted with static analysis.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,14 @@ This library is tested against:
   
 ## Node.js Support
 
-`grpc-web-client` also [supports Node.js through a transport](ts/docs/transport.md#node-http-only-available-in-a-nodejs-environment) that uses the `http` and `https` packages. Usage does not vary from browser usage as transport is determined at runtime.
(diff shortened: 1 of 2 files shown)


jonahbron:
I've been trying to get all of the integration tests passing again, but it's not cooperating.  I moved the same exact Nodejs detection into the unit test file, but it's not working as expected.  The only job that fails because of my changes is the Node.js one.  The others are just timeouts, I assume the test took too long.

https://travis-ci.org/improbable-eng/grpc-web/builds/389948600

johanbrandhorst:
Having thought about it, we're still in the 0.x range of semantic versioning so I think a breaking change could be alright (going to 0.7).

easyCZ:
As part of this, we may want to consider adding support for setting of the default transport. This would eliminate the need to always explicitly override the transport in Node. On the other hand, it's easy to wrap by the client to auto-add the node-transport so I'm not too fussed.

Issue #111, opened 2017-12-13
Unable to access /metrics or /debug/events on grpcwebproxy


cristipp:
Using Firefox. Serve port on 8081, debug port on 8082.
```
/go/bin/grpcwebproxy \
    --server_tls_cert_file=grpcwebproxy.crt \
    --server_tls_key_file=grpcwebproxy.key \
    --server_http_tls_port=8081 \
    --server_http_debug_port=8082 \
    --server_tls_client_cert_verification=none \
    --backend_tls_noverify \
    --backend_addr=localhost:8080
```

`http://localhost:8081/metrics` => `�`
`https://localhost:8081/metrics` => `invalid gRPC request method`
`http://localhost:8082/metrics` => `gRPC requires HTTP/2`
`https://localhost:8082/metrics` => `An error occurred during a connection to localhost:8082. SSL received a record that exceeded the maximum permissible length. Error code: SSL_ERROR_RX_RECORD_TOO_LONG`

easyCZ:
@cristipp Is this still a problem? Have you found a work around or the source of the problem?

AminLA:
@easyCZ 
this can be fixed by switch from grpc server (w.server) to http.DefaultServeMux

```go
// You can control the CORS behaviour using `With*` options in the WrapServer function.
func (w *WrappedGrpcServer) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
	if w.IsAcceptableGrpcCorsRequest(req) || w.IsGrpcWebRequest(req) {
		w.corsWrapper.Handler(http.HandlerFunc(w.HandleGrpcWebRequest)).ServeHTTP(resp, req)
		return
	}   
    // debug/events, debug/requests can not handle by grpc server, use DefaultServeMux instead
    //w.server.ServeHTTP(resp, req)
    http.DefaultServeMux.ServeHTTP(resp, req)
}
```

n0mer:
@easyCZ any update on this?

easyCZ:
Looking into this. Ideally the solution would both solve this issue and usage of the `grpc-web` wrapper as middlewares for chi & echo - https://github.com/easyCZ/grpc-web-hacker-news/blob/master/server/middleware/grpcWeb.go

sandersaares:
This remains an issue in latest master branch.

PR #208, opened 2018-06-14
Fix serving of Promethius metrics

amerry:
Using http.Handle will not work because the server is not using http.DefaultServeMux.

This is a minimal and naive fix for #111. It is still only installed on the non-TLS port, both to limit its exposure on production systems and because of the possibility of interference with the proxying behaviour.


amerry:
I think this version of the fix is cleaner, and also enables pprof to be accessed (after all, someone has imported pprof, presumably in the expectation it will be available).

PR #213, opened 2018-06-25
Proxy: Correct Content-Type in response

johanbrandhorst:
The Content-Type header in the request is automatically changed from application/grpc-web to application/grpc, however, no corresponding conversion is done for response headers and trailers. This commit adds this functionality, both to early headers and to the in-body trailers.

Found while writing my own WASM gRPC-Web client in Go (in a fork of grpc/grpc-go).

### DIFF
--- a/go/grpcweb/grpc_web_response.go
+++ b/go/grpcweb/grpc_web_response.go
@@ -90,6 +90,7 @@ func (w *grpcWebResponse) copyTrailersAndHeadersToWrapped() {
 		}
 	}
 	w.writeCorsExposedHeaders()
+	hackIntogRPCWebContentType(wrappedHeader)
 	w.wrapped.WriteHeader(http.StatusOK)
 	w.wrapped.(http.Flusher).Flush()
 }
@@ -134,5 +135,16 @@ func (w *grpcWebResponse) extractTrailerHeaders() trailer {
 			trailerHeaders.Add(k, v)
 		}
 	}
+	hackIntogRPCWebContentType(trailerHeaders)
 	return trailerHeaders
+}
+
+type headerLike interface {
+	Get(string) string
+	Set(string, string)
+}
+
+func hackIntogRPCWebContentType(in headerLike) {
+	contentType := in.Get("content-type")
+	in.Set("content-type", strings.Replace(contentType, "application/grpc", "application/grpc-web", 1))
 }
--- a/go/grpcweb/trailer.go
+++ b/go/grpcweb/trailer.go
@@ -24,9 +24,15 @@ func (t trailer) Get(key string) string {
 	if t.Header == nil {
 		return ""
 	}
+	key = strings.ToLower(key)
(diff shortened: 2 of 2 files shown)


johanbrandhorst:
@MarcusLongmuir @easyCZ I hope this should be a pretty easy merge. It should be noted that frontend clients still see `application/octet-stream` for certain streamed responses, but this is a best effort to correct the content type where possible.

johanbrandhorst:
For reference see https://github.com/grpc/grpc-go/pull/2174

johanbrandhorst:
Bump. Build failures all look like flaky tests to me.

johanbrandhorst:
_Bump_.

johanbrandhorst:
Bump

domgreen:
@johanbrandhorst  i think the reason its not been merged is due to the build break :( will reach out to @easyCZ and @MarcusLongmuir internally. 👍

johanbrandhorst:
Can I have a rerun with the new CI fixes?

MarcusLongmuir:
@johanbrandhorst: Can you rebase on top of latest master please?

johanbrandhorst:
@MarcusLongmuir doh :man_facepalming: should've done that already. Thanks!

MarcusLongmuir:
@johanbrandhorst: The failing tests look related to headers. This might not be flaky tests.

johanbrandhorst:
Thanks for looking @MarcusLongmuir, could it be we need to adapt the client as well? It seems to only fail in some browsers still though so presumably it's browser specific behaviour. Any ideas where the error might be?

johanbrandhorst:
The more I test the `grpcwebproxy` versus the `envoy` proxy the more convinced I am the `grpcwebproxy` is breaking the spec here (ref: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md#protocol-differences-vs-grpc-over-http2). Regardless of what the tests say, we should be returning `application/grpc-web` from the proxy. Is there some reason the client wouldn't work with this? I'd like to get this merged

johanbrandhorst:
See https://github.com/johanbrandhorst/grpc-web-compatibility-test for more information.

johanbrandhorst:
It seems to me the failing tests are IE and some versions of Safari - this PR is only changing `application/grpc` to `application/grpc-web`. It doesn't seem like we need to do any work in the clients, but I don't know why the tests are failing.

jonny-improbable:
Morning @johanbrandhorst just stumbled across this PR - there's been a lot of change in the repo over the last month; could you rebase and we can give it another spin through CI :)

johanbrandhorst:
Yeah, I'll give it another rebase, hoping for the best!

johanbrandhorst:
Still seems to be failing edge tests and one safari test.

johanbrandhorst:
Gonna close this as it's very old and will probably be solved by whatever we do about #334

Issue #225, opened 2018-08-27
gRPC status code header destroyed on error


hectim:
Been using this grpc-web client on a project for a while, but recently started handling specific grpc errors on the front-end. Our requests hit an Envoy proxy, which does its own httpToGrpc status conversion, but in the grpc-web-client code specifically we are also losing the grpc-status header. 

Through some local testing, it looks like the problem comes from https://github.com/improbable-eng/grpc-web/blob/master/ts/src/client.ts#L99
```
const code = httpStatusToCode(status);
this.props.debug && debug("onHeaders.code", code);
const gRPCMessage = headers.get("grpc-message") || [];
this.props.debug && debug("onHeaders.gRPCMessage", gRPCMessage);
if (code !== Code.OK) {
  const statusMessage = this.decodeGRPCStatus(gRPCMessage[0]);
  this.rawOnError(code, statusMessage);
  return;
}
```

My problem could be solved if instead of passing the newly created `code` to rawOnError, it checked for a grpc-status header first and used that value before defaulting to the new `code`.  Something like: 
```
const code = parseInt(headers.get("grpc-status"), 10) || httpStatusToCode(status);
```
This seems to be the simplest way to fix it, although there may be reasons why this is unwanted. 

Option 2: If we made sure the headers were being passed correctly, we could just pull them off the response. Right now when it throws an error it makes a `new Metadata()` and hands that off, but it that isn't very helpful. 
```
rawOnError(code: Code, msg: string) {
    this.props.debug && debug("rawOnError", code, msg);
    if (this.completed) return;
    this.completed = true;
    this.onEndCallbacks.forEach(callback => {
      detach(() => {
        callback(code, msg, new Metadata());
      });
    });
  }
```

If instead, rawOnError took an optional third parameter of the headers 
```
rawOnError(code: Code, msg: string, headers?: Metadata) {
```
and passed those (if they exist) to the callback, we could accomplish essentially the same thing. 
```
callback(code, msg, headers || new Metadata());
```

I could submit a PR for this anytime, just wanted to know if there were any extra thoughts about potential hiccups or problems with either of these. Personally, I would opt for the first option for simplicity.

gunn4r:
@easyCZ I'm working with @hectim on this and will be putting together the actual PR. Thanks a bunch!

gunn4r:
@easyCZ After digging through this more, it appears the tests that are testing "error" cases are not actually ever hitting the error case here:

```if (code !== Code.OK) {
[...]

PR #226, opened 2018-08-29
Fixes headers/trailers being lost when grpc-status is non-zero. 

gunn4r:
- rawOnHeaders now called even on grpc errors.
- Derived grpc status `code` defaults to grpc-status if available on headers with `httpStatusToCode()` as fallback.
- headers (trailers) now provided to rawOnError (defaults to new Metadata () if none provided).

Closes issue #225

### DIFF
--- a/ts/src/client.ts
+++ b/ts/src/client.ts
@@ -100,7 +100,7 @@ class GrpcClient<TRequest extends ProtobufMessage, TResponse extends ProtobufMes
       const gRPCStatus: Code = parseInt(headers.get("grpc-status")[0], 10);
       this.props.debug && debug("onHeaders.gRPCStatus", gRPCStatus);
 
-      const code: Code = gRPCStatus >= 0 ? gRPCStatus : httpStatusToCode(status);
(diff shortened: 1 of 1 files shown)


ts/src/client.ts
```diff
@@ -219,13 +224,13 @@ class GrpcClient<TRequest extends ProtobufMessage, TResponse extends ProtobufMes
     });
   }
 
-  rawOnError(code: Code, msg: string) {
+  rawOnError(code: Code, msg: string, trailers: Metadata = new Metadata()) {
```
coltonmorris: Aren't these headers and not trailers?

gunn4r: Trailers in fact!

rawOnError calls the `onEndCallbacks` (https://github.com/improbable-eng/grpc-web/blob/master/ts/src/client.ts#L222)

Which has this signature:
`  onEndCallbacks: Array<(code: Code, message: string, trailers: Metadata) => void> = [];`
https://github.com/improbable-eng/grpc-web/blob/master/ts/src/client.ts#L44

PR #237, opened 2018-09-09
Minor fixes

johanbrandhorst:
1. Install stable version of `dep` in CI.
1. Regenerate Gopkg.lock with new `dep`.
1. Add `protoc-gen-go` as explicit dependency.
1. Install dep locked version of `protoc-gen-go` in CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 go:
-  - 1.8
+  - "1.11"
 go_import_path: github.com/improbable-eng/grpc-web
 addons:
   hosts:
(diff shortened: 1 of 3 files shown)


johanbrandhorst:
On second thought, it seems like this will run about 30 tests for all 4 support architectures. Do we want to support a single architecture in CI instead? I'm hesitant to leave it at 1.8.

easyCZ:
Good spot. We could bump it to 1.10. Running on multiple go versions will result in even more flakes. I'd be inclined to run just on one and perhaps attempt to find a path forward to improve reliability of the tests

johanbrandhorst:
Any reason not to change it to 1.11? Is Improbable still running 1.10 internally ;)? I agree we should run just 1 version but I'd be inclined to run latest if possible.

easyCZ:
We are indeed on 1.10 still but it should not matter for us internally as we can upgrade when needed.

Yeah let's go with 1.11 then.

PR #244, opened 2018-09-14
Switched to assign CAs to correct config property

### LINKED ISSUE #123: proxy: certificate validation of backend service not working
If you specify a CA certificate to use in the proxy to verify the identity of the backend service, the certificate will en up as entry in tlsConfig.ClientCAs but the godocs are stating that:

```
// ClientCAs defines the set of root certificate authorities
// that servers use if required to verify a client certificate
// by the policy in ClientAuth.
```

So the CA-certificate should be installed there when preparing a server side tlsConfig. When preparing a client side config it should be placed in RootCAs instead (so the fix is trivial).

I think the behaviour has become more strictly with newer go-versions. I think it used to work.

jonahbron:
Resolves #123

### DIFF
--- a/go/grpcwebproxy/backend.go
+++ b/go/grpcwebproxy/backend.go
@@ -62,13 +62,13 @@ func buildBackendTlsOrFail() *tls.Config {
 		tlsConfig.InsecureSkipVerify = true
 	} else {
 		if len(*flagBackendTlsCa) > 0 {
-			tlsConfig.ClientCAs = x509.NewCertPool()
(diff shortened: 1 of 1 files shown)


johanbrandhorst:
I don't know this area of the code very well, would it be possible to add a test for this?

jonahbron:
@johanbrandhorst I'm not set up to really write Go on my machine, and I see no existing testing infrastructure for the proxy, so I wouldn't actually know how to set that up (don't normally write Go).

johanbrandhorst:
I've restarted the failing test

Issue #245, opened 2018-09-15
Add option to Increase Max Message Size on the Proxy


nevi-me:
Is there a config option to increase max message size on the proxy?

There's a similar issue #21, but this one was extreme as I only need a few more MB on my message.

I'm not familiar with Go, but I know that there is a dial option to increase message size: https://github.com/grpc/grpc-go/blob/master/dialoptions.go#L147. 

I'm wondering if this has been exposed on the grpc-web proxy?

johanbrandhorst:
I don't think this is exposed, but there's no reason we couldn't add such a setting. Would you be interested in contributing this?

johanbrandhorst:
As a separate point, the 4MB message limit is already pretty high - personally I would switch to streaming if the messages are larger than 1MB.

nevi-me:
I know it's high, and I tried streaming, but I end up streaming about 2000 small messages, to which the proxy throws an error after between 300-500 messages in my stream. I didn't capture the logs at the time, but I'll try reproduce it.

johanbrandhorst:
Hmm yeah that sounds like a bug worth investigating.

nevi-me:
I steam the ~2000 records from MongoDB using a NodeJS stream, then for `stream.on('data', {})` I pass the message to my stream.

It could be that the errors are because of the stream being saturated. Given that my stuff's dependent on a lot of things, I might struggle with creating a standalone repro that works on the streaming bit, but I'll try (and open a separate issue if I succeed).

I had a look at the proxy's Go code with where the flags are set. I'll try adding the option to increase max size, but I might need some hand-holding if you (and anyone else) don't mind.

johanbrandhorst:
Of course, if you put a PR up I'll review it :).

nevi-me:
How do I overcome `cannot find package "google.golang.org/grpc/transport"` error? I've come across the issue before when I was setting up the proxy on my environment, but I can't remember what workaround I applied, or where I found the source of that 😫

johanbrandhorst:
@nevi-me You'll see [in my reply on #216](https://github.com/improbable-eng/grpc-web/issues/216#issuecomment-420189052) that this is only an issue if you try to build the proxy without using the appropriate dependency.

johanbrandhorst:
Fixed with #246

PR #246, opened 2018-09-15
Add option to increase max message size

nevi-me:
Adds the option to pass an integer value as the `grpc.WithMaxMsgSize` from https://godoc.org/google.golang.org/grpc#WithMaxMsgSize, https://github.com/grpc/grpc-go/blob/master/dialoptions.go#L147.

ref #245

### DIFF
--- a/go/grpcwebproxy/backend.go
+++ b/go/grpcwebproxy/backend.go
@@ -54,7 +54,7 @@ func dialBackendOrFail() *grpc.ClientConn {
 	} else {
 		opt = append(opt, grpc.WithInsecure())
 	}
-	opt = append(opt, grpc.MaxCallRecvMsgSize(*flagMaxCallRecvMsgSize))
+	opt = append(opt, grpc.WithDefaultCallOptions(MaxCallRecvMsgSize(*flagMaxCallRecvMsgSize)))
 	cc, err := grpc.Dial(*flagBackendHostPort, opt...)
 	if err != nil {
 		logrus.Fatalf("failed dialing backend: %v", err)
--- a/go/grpcwebproxy/backend.go
+++ b/go/grpcwebproxy/backend.go
@@ -54,7 +54,7 @@ func dialBackendOrFail() *grpc.ClientConn {
 	} else {
 		opt = append(opt, grpc.WithInsecure())
 	}
-	opt = append(opt, grpc.WithDefaultCallOptions(MaxCallRecvMsgSize(*flagMaxCallRecvMsgSize)))
+	opt = append(opt, grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(*flagMaxCallRecvMsgSize)))
 	cc, err := grpc.Dial(*flagBackendHostPort, opt...)
 	if err != nil {
 		logrus.Fatalf("failed dialing backend: %v", err)
--- a/go/grpcwebproxy/backend.go
+++ b/go/grpcwebproxy/backend.go
@@ -55,7 +55,7 @@ func dialBackendOrFail() *grpc.ClientConn {
 		opt = append(opt, grpc.WithInsecure())
 	}
 	if *flagMaxMsgSize != 4194304 {
-		opt = append(opt, grpc.WithMaxMsgSize())
+		opt = append(opt, grpc.WithMaxMsgSize(flagMaxMsgSize))
 	}
 	cc, err := grpc.Dial(*flagBackendHostPort, opt...)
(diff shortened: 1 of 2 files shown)


go/grpcwebproxy/backend.go
```diff
@@ -30,6 +30,12 @@ var (
 		"Whether to ignore TLS verification checks (cert validity, hostname). *DO NOT USE IN PRODUCTION*.",
 	)
 
+	flagMaxMsgSize = pflag.Int(
+		"backend_max_msg_size",
+		4194304,
```
johanbrandhorst: Can we make this equivalent to the current max? https://github.com/grpc/grpc-go/blob/v1.8.2/server.go#L54. Also please add a reference to this line on github (i.e., copy this link). Note that the link is for v1.8.2 which is the version we've locked in Gopkg.toml.

go/grpcwebproxy/backend.go
```diff
@@ -48,6 +54,9 @@ func dialBackendOrFail() *grpc.ClientConn {
 	} else {
 		opt = append(opt, grpc.WithInsecure())
 	}
+	if *flagMaxMsgSize != 4194304 {
```
johanbrandhorst: This check is unecessary - just always add the option.

go/grpcwebproxy/backend.go
```diff
@@ -30,6 +30,12 @@ var (
 		"Whether to ignore TLS verification checks (cert validity, hostname). *DO NOT USE IN PRODUCTION*.",
 	)
 
+	flagMaxMsgSize = pflag.Int(
+		"backend_max_msg_size",
+		4194304,
+		"Maximum message size limit. If not specified, the default of 4MB will be used.",
```
johanbrandhorst: _receive_ size limit.

go/grpcwebproxy/backend.go
```diff
@@ -30,6 +30,12 @@ var (
 		"Whether to ignore TLS verification checks (cert validity, hostname). *DO NOT USE IN PRODUCTION*.",
 	)
 
+	flagMaxMsgSize = pflag.Int(
+		"backend_max_msg_size",
```
johanbrandhorst: Lets be explicit about this being the max _receive_ message size.

go/grpcwebproxy/backend.go
```diff
@@ -48,6 +54,9 @@ func dialBackendOrFail() *grpc.ClientConn {
 	} else {
 		opt = append(opt, grpc.WithInsecure())
 	}
+	if *flagMaxMsgSize != 4194304 {
+		opt = append(opt, grpc.WithMaxMsgSize(flagMaxMsgSize))
```
johanbrandhorst: This call option is deprecated - please use `MaxCallRecvMsgSize` (https://github.com/grpc/grpc-go/blob/v1.8.2/rpc_util.go#L215).

go/grpcwebproxy/backend.go
nevi-me: I thought the max is 4MB, I'll update

nevi-me:
Apologies, I still struggled to compile the binary after `dep ensure`, so I didn't see that I was passing the wrong type.

I've also added some documentation for the option on the README, so we can take care of that while we're at it.

johanbrandhorst:
Let's just see how the build goes but the code LGTM.

nevi-me:
I might not be able to immediately use the updated binary, my go won't compile, errors with $GOPATH. Do you know if Travis generates build artifacts?

johanbrandhorst:
I don't think Travis does. It really should not be hard to build if you have go installed already. Have you checked the README in the grpcwebproxy folder? It has a step by step guide.

johanbrandhorst:
Thanks for your contribution!

nevi-me:
I got dep installed, but when I `go install ./go/grpcwebproxy`, go complains about my $GOPATH. I'll have a look at it tomorrow, I'm retiring for the night (GMT +2). Thanks for holding my hand through the process Johan!

johanbrandhorst:
:grimacing: this is probably easier to resolve in a chat: there's a gRPC-Web slack you can join and we can debug your setup in there: https://join.slack.com/t/improbable-eng/shared_invite/enQtMzQ1ODcyMzQ5MjM4LWY5ZWZmNGM2ODc5MmViNmQ3ZTA3ZTY3NzQwOTBlMTkzZmIxZTIxODk0OWU3YjZhNWVlNDU3MDlkZGViZjhkMjc

PR #253, opened 2018-10-05
Prevent leaking goroutines from websocket connections

amerry:
The write to the CloseNotify() channel was blocking, preventing the
Read() call from exiting, and preventing ServeHTTP from returning. This
meant that requests tended to stay around forever in a deadlocked state.

Using the context package prevent this, as the cancellation is
non-blocking. It depends on Go 1.7.

Note that the writer still needs to support the CloseNotifier interface,
otherwise other things break, but it does not seem to be necessary to
signal using it.

### DIFF
--- a/go/grpcweb/wrapper.go
+++ b/go/grpcweb/wrapper.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"net/url"
 
+	"context"
 	"strings"
 	"time"
 
@@ -148,14 +149,17 @@ func (w *WrappedGrpcServer) handleWebSocket(wsConn *websocket.Conn, req *http.Re
 		return
 	}
 
+	ctx, cancelFunc := context.WithCancel(req.Context())
+	defer cancelFunc()
+
 	respWriter := newWebSocketResponseWriter(wsConn)
-	wrappedReader := newWebsocketWrappedReader(wsConn, respWriter)
+	wrappedReader := newWebsocketWrappedReader(wsConn, respWriter, cancelFunc)
 
 	req.Body = wrappedReader
 	req.Method = http.MethodPost
 	req.Header = headers
 
-	w.server.ServeHTTP(respWriter, hackIntoNormalGrpcRequest(req))
+	w.server.ServeHTTP(respWriter, hackIntoNormalGrpcRequest(req.WithContext(ctx)))
 }
 
 // IsGrpcWebRequest determines if a request is a gRPC-Web request by checking that the "content-type" is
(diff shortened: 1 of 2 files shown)


go/grpcweb/wrapper.go
```diff
@@ -148,14 +149,17 @@ func (w *WrappedGrpcServer) handleWebSocket(wsConn *websocket.Conn, req *http.Re
 		return
 	}
 
+	ctx, cancelFunc := context.WithCancel(req.Context())
+	defer cancelFunc()
```
johanbrandhorst: Do we need to explicitly cancel this? Can we not rely on the built-in request context being cancelled?

johanbrandhorst: If we can rely on the request context we don't need to pass around a cancelFunc either. I assume you've tested some of this?

amerry: Yeah, we've got a websocket client that makes a request every second, so the memory leak is really obvious, and I put in some logging for testing (tracking the start and end of handleWebSocket, as well as before and after the cancellation step in the reader).

I tested a bunch of different configurations. Not cancelling the request context also prevented connections from closing properly; I'm not sure why (you would expect `w.server.ServeHTTP` to pick up from the error returned from the `Body` that it would need to cancel the request, but this doesn't seem to happen).

I'll admit that I didn't go diving into the lower-level code to see what was going on - this was mostly by trial-and-error.

johanbrandhorst: OK - I'm not worried that this will cause any real problems, since the request is over once this function has exited, just thought it was a bit overzealous, but if your testing implies the necessity, I'm OK with it!

johanbrandhorst:
Would like a confirmation from @MarcusLongmuir that we're OK requiring Go 1.7+ for building this now.

amerry:
Oh, wait, I think I misunderstood. That `defer` there isn't necessary to prevent the goroutine leaking. I just put it in based on the `context.WithCancel` documentation, which seemed to suggest it was good practice.

johanbrandhorst:
Let me rephrase: can you categorically confirm that _just_ removing the channel doesn't work? I imagine the request context will be cancelled when the client finishes reading.

johanbrandhorst:
Going to give the last error another go :grimacing:

johanbrandhorst:
Just going to give @MarcusLongmuir and @easyCZ a couple of days to put anymore comments on this but LGTM!

johanbrandhorst:
OK I think we've had enough time to think about this, going to merge, thanks for your contribution!

PR #255, opened 2018-10-09
Fix git clone path

seregayoga:
Git clone doesn't work for me
```
$ git clone github.com/improbable-eng/grpc-web $GOPATH/src/
fatal: repository 'github.com/improbable-eng/grpc-web' does not exist
```
So I changed the path.

### DIFF
--- a/go/grpcwebproxy/README.md
+++ b/go/grpcwebproxy/README.md
@@ -27,7 +27,7 @@ https://github.com/improbable-eng/grpc-web/releases/tag/0.6.3
 To build, you need to have Go >= 1.8, and call `go get` with `dep ensure`:
 
 ```sh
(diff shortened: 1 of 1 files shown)


johanbrandhorst:
Hm, I'm not sure we should use Https cloning by default. I'd rather this document just said "clone the repo into the GOPATH" and the user can figure out what that means maybe?

seregayoga:
> I'm not sure we should use Https cloning by default.
Why not? On the other hand, maybe someone wants to install binary quickly by copying and pasting only this four lines.

johanbrandhorst:
The instructions in this README are for developers. If you want to download the binary we have release builds for most platforms. I think we should of course correct the error if the command doesn't work, but it's not targeted at users who just want to run the proxy.

PR #258, opened 2018-10-16
client: don't call callbacks if close() is called

### LINKED ISSUE #257: race condition on stream close
This is for a server-side stream. The stack trace I'm getting on the client is something like so:

Uncaught TypeError: Cannot read property 'data' of null

* [`onMessage`](https://github.com/improbable-eng/grpc-web/blob/1464bd937bd627adfbc086a857cd3d917eb63d29/example/ts/_proto/examplecom/library/book_service_pb_service.js#L73)
* `Array.<anonymous>`
* [`runCallbacks`](https://github.com/improbable-eng/grpc-web/blob/72eb701d6f320ca324b3347c7925a720b553eae5/ts/src/detach.ts#L11)

This occurs when I'm clicking on my page rapidly, causing the stream to be opened/closed over and over again. I haven't tried to reproduce it using the example project.

It seems to me that perhaps the [setTimeout(0) in `detach.ts`](https://github.com/improbable-eng/grpc-web/blob/72eb701d6f320ca324b3347c7925a720b553eae5/ts/src/detach.ts#L40) is the cause of the race condition, as there doesn't seem to be a mechanism to remove callbacks that are queued via the timeout.

One workaround would be for `ts-protoc-gen` to add a check to see if listeners is `null` in the generated `onMessage` callbacks. I suspect a similar race condition exists in `onEnd` also, which could be fixable via the same workaround.

virtuald:
- This could break some existing clients using the raw grpc API (maybe should document it?), but ts-protoc-gen clears out the callbacks anyways so this shouldn't cause anyone pain
- Fixes #257

... I'd write a test for this, but I'm not actually a Javascript/Typescript developer so I'm not familiar with the test framework.

### DIFF
--- a/ts/src/client.ts
+++ b/ts/src/client.ts
@@ -209,6 +209,7 @@ class GrpcClient<TRequest extends ProtobufMessage, TResponse extends ProtobufMes
 
     this.onEndCallbacks.forEach(callback => {
       detach(() => {
+        if (this.closed) return;
         callback(code, message, trailers);
       });
     });
@@ -230,6 +231,7 @@ class GrpcClient<TRequest extends ProtobufMessage, TResponse extends ProtobufMes
     this.completed = true;
     this.onEndCallbacks.forEach(callback => {
       detach(() => {
+        if (this.closed) return;
         callback(code, msg, trailers);
       });
     });
@@ -240,6 +242,7 @@ class GrpcClient<TRequest extends ProtobufMessage, TResponse extends ProtobufMes
     if (this.completed || this.closed) return;
     this.onMessageCallbacks.forEach(callback => {
       detach(() => {
+        if (this.closed) return;
         callback(res);
       });
     });


virtuald:
Worth noting that when I built this PR into my application, the errors I was experiencing went away.

virtuald:
Ping.

jonnyreeves:
Just had a look into the build failure, I believe the root cause is the failure to bring up Chrome:

```
Error: [init({"project":"master","build":"travis_803","acceptSslCerts":true,"defaultVideo":true,"browserstack.local":true,"browserstack.tunnel":true,"browserstack.debug":true,"tunnelIdentifier":"tunnel-0.9422959229946926","browserstack.localIdentifier":"tunnel-0.9422959229946926","testSuite":"client","browserName":"chrome","browserVersion":"42","os":"Windows","os_version":"7"})] The environment you requested was unavailable.
Could not start Browser / Emulator
```

We are seeing the same failure on master so I am also confident this failure is unrelated to the changes.

I'll merge this through, however we should follow up with fixing this particular browser test, or remove it from the matrix

jonny-improbable:
@virtuald could you please bring this up to speed with master now #262 was merged? Thanks!

virtuald:
Looks like it passed. 👍

johanbrandhorst:
LGTM!

Issue #103, opened 2017-11-22
Option to set xhr.withCredentials


aakoshh:
I'm facing a scenario where I have to talk to a gRPC server that sits behind a reverse proxy which redirects to ADFS for login and sets a http-only session cookie which it will pick up on subsequent requests and attaches an `Authorization` header. The problem is that the transports don't set the cookies with the gRPC invocation, so each POST results in a redirect. 

What do you think about adding an option somewhere which could set the `withCredentials` property of the `XMLHttpRequest` before sending the request?

fabxc:
Looks like the CORS headers set for OPTIONS requests specifically set `Access-Control-Allow-Credentials: true` but the JS/TS implementation hard-codes the opposite (e.g. [here](https://github.com/improbable-eng/grpc-web/blob/master/ts/src/transports/fetch.ts#L37) not allowing to set `withCredentials = true` in the `XMLHTTPRequest`.

This seems like an unnecessary restriction that forces annoying constraints on the rest of the setup. Providing a fully custom transport for this seems like a lot of overhead.

jonnyreeves:
So while we can solve this problem for fetch and xhr based transports, there is nothing that can be done for including cookies on cross-origin WebSocket connections as far as I know.

One workaround would be to make the WebSocketTransport throw an error if you try to construct it in such a way that you request credentials (cookies) be send cross origin (`credentials: 'include'`).

MOZGIII:
Why not let the browser deal with that?
I was looking for a way to pass `credentials: 'include'` to the `fetch` transport - and found this issue. Our usecase is actually sending the corss-origin request with cerdentials - it's allowd by fetch spec.

MOZGIII:
We ended up copying `fetch` transport from tree to our source code and patching it so it passes `include`.
In general it would be great to have an ability to configure various transport-specific parameters if we opt out of transport autodetection by pass transport manually. Seems like a good idea to me - it doesn't require a generic interface deduction of any kind, and just provides a flexible way of using that's already in there.

I can compose a sample PR if you like this idea.

johanbrandhorst:
Sounds good to me, let's get a PR up for discussion.

MOZGIII:
@johanbrandhorst please take a look at #260

MOZGIII:
And #261.

jonny-improbable:
This has been resolved for some time now, you can now configure the `CrossBrowserTransport`.

maxmcd:
[...]

PR #260, opened 2018-10-19
Transport options (variant A)

MOZGIII:
See #103 

What I don't like about this it that holds the transport specific options in the generic structure, although you might like it, so I decided to roll it out for discussion.

I'll try to compose a variant B of the similar idea, stay tuned.

This is conflicting implementation to #261.

This variant is incomplete, tests don't pass, this is just a rough idea of how it can be done.

### DIFF
--- a/ts/src/transports/Transport.ts
+++ b/ts/src/transports/Transport.ts
@@ -14,21 +14,22 @@ export interface Transport {
   start(metadata: Metadata): void
 }
 
-export interface TransportConstructor {
-  (options: TransportOptions): Transport | Error;
+export interface TransportConstructor<TImplementaionSpecificOptions= undefined> {
+  (options: TransportOptions<TImplementaionSpecificOptions>): Transport | Error;
 }
 
-export interface TransportOptions {
+export interface TransportOptions<TImplementaionSpecificOptions = undefined> {
   methodDefinition: MethodDefinition<ProtobufMessage, ProtobufMessage>;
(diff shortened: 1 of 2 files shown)


ts/src/transports/Transport.ts
```diff
@@ -14,21 +14,22 @@ export interface Transport {
...
 }
 
-export interface TransportConstructor {
-  (options: TransportOptions): Transport | Error;
+export interface TransportConstructor<TImplementaionSpecificOptions= undefined> {
+  (options: TransportOptions<TImplementaionSpecificOptions>): Transport | Error;
 }
 
-export interface TransportOptions {
+export interface TransportOptions<TImplementaionSpecificOptions = undefined> {
```
johanbrandhorst:
```suggestion
export interface TransportOptions<TImplementationSpecificOptions = undefined> {
```

ts/src/transports/Transport.ts
```diff
@@ -14,21 +14,22 @@ export interface Transport {
...
   methodDefinition: MethodDefinition<ProtobufMessage, ProtobufMessage>;
   debug: boolean;
   url: string;
+  transportSpecificOptions: TImplementaionSpecificOptions;
```
johanbrandhorst:
```suggestion
  transportSpecificOptions: TImplementationSpecificOptions;
```

MOZGIII:
Moving on to variant B

PR #261, opened 2018-10-19
Transport options (variant B)

MOZGIII:
See #103 
This better captures the idea I poposed at #103. It's also simpler than #260 and users have to opt-in to specify custom tranport.

### DIFF
--- a/test/ts/src/testRpcCombinations.ts
+++ b/test/ts/src/testRpcCombinations.ts
@@ -3,7 +3,6 @@ import {
   corsHost
 } from "../../hosts-config";
 import {grpc} from "../../../ts/src/index";
-import { fetchRequestWithOptions } from "../../../ts/src/transports/fetch";
 
 type TestConfig = {
   testHostUrl: string,
@@ -59,7 +58,7 @@ export function runWithSupportedTransports(cb: (transport: grpc.TransportConstru
   };
 
   if (typeof window !== "undefined") {
-    transports["fetchWithOptions"] = fetchRequestWithOptions({})
+    transports["fetchWithOptions"] = grpc.transports.fetchRequestWithOptions({})
   }
 
   if (!process.env.DISABLE_WEBSOCKET_TESTS) {
--- a/test/ts/src/testRpcCombinations.ts
+++ b/test/ts/src/testRpcCombinations.ts
@@ -3,6 +3,7 @@ import {
   corsHost
 } from "../../hosts-config";
 import {grpc} from "../../../ts/src/index";
+import { fetchRequestWithOptions } from "../../../ts/src/transports/fetch";
 
 type TestConfig = {
   testHostUrl: string,
@@ -54,7 +55,8 @@ export function runWithHttp1AndHttp2(cb: (config: TestConfig) => void) {
 
 export function runWithSupportedTransports(cb: (transport: grpc.TransportConstructor | undefined) => void) {
   const transports: {[key: string]: grpc.TransportConstructor | undefined} = {
-    "defaultTransport": undefined
+    "defaultTransport": undefined,
+    "fetchWithOptions": fetchRequestWithOptions({}),
   };
 
   if (!process.env.DISABLE_WEBSOCKET_TESTS) {
--- a/test/ts/src/testRpcCombinations.ts
+++ b/test/ts/src/testRpcCombinations.ts
@@ -55,9 +55,12 @@ export function runWithHttp1AndHttp2(cb: (config: TestConfig) => void) {
 
 export function runWithSupportedTransports(cb: (transport: grpc.TransportConstructor | undefined) => void) {
   const transports: {[key: string]: grpc.TransportConstructor | undefined} = {
-    "defaultTransport": undefined,
-    "fetchWithOptions": fetchRequestWithOptions({}),
+    "defaultTransport": undefined
   };
+
+  if (typeof window !== "undefined") {
+    transports["fetchWithOptions"] = fetchRequestWithOptions({})
+  }
 
   if (!process.env.DISABLE_WEBSOCKET_TESTS) {
     transports["websocketTransport"] = grpc.WebsocketTransportFactory
--- a/ts/src/transports/fetch.ts
+++ b/ts/src/transports/fetch.ts
@@ -3,10 +3,14 @@ import {Transport, TransportOptions} from "./Transport";
 import {debug} from "../debug";
 import detach from "../detach";
 
+interface FetchOptions {
+  credentials?: RequestCredentials;
+}
+
 /* fetchRequest uses Fetch (with ReadableStream) to read response chunks without buffering the entire response. */
-export default function fetchRequest(options: TransportOptions): Transport {
-  options.debug && debug("fetchRequest", options);
-  return new Fetch(options);
+export default function fetchRequest(options: TransportOptions, fetchOptions: FetchOptions = {}): Transport {
+  options.debug && debug("fetchRequest", options, fetchOptions);
+  return new Fetch(options, fetchOptions);
 }
 
 declare const Response: any;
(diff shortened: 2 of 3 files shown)


ts/src/transports/fetch.ts
```diff
@@ -3,10 +3,21 @@ import {Transport, TransportOptions} from "./Transport";
...
+  return function(options: TransportOptions) {
+    return fetchRequest(options, fetchOptions);
+  }
+}
+
 /* fetchRequest uses Fetch (with ReadableStream) to read response chunks without buffering the entire response. */
-export default function fetchRequest(options: TransportOptions): Transport {
-  options.debug && debug("fetchRequest", options);
-  return new Fetch(options);
+export default function fetchRequest(options: TransportOptions, fetchOptions: FetchOptions = {}): Transport {
```
johanbrandhorst: Do we need both of these functions? It seems like a bit of duplication to me. Can you quantify the benefit of having both? Personally I think maybe just this one would suffice.

MOZGIII: The `fetchRequest` is preserved (but extended with additional optional argument) to keep the existing API intact.

Usage with the API before this PR (may not precise code):

```typescript
grpc.invoke({
  // ...
  transport: fetchRequest
})
```

Notice `fetchRequest` is passed as value, and is not called directly. This means there's no way to pass that additional argument - as it's called internally by the library.

With the new `fetchRequestWithOptions` we'd use it like this:

```typescript
grpc.invoke({
  // ...
  transport: fetchRequestWithOptions({ credentials: "include" })
})
```

This effectively creates a carried version of the `fetchRequest` with the options we need to pass, allowing the desired usecase.

I'm not against changing any of that, I just wanted this PR to not break the existing APIs.

johanbrandhorst: So `fetchRequestWithOptions` is required to access this new API? Of course, not breaking existing API is essential. I'm just showing my lack of TypeScript experience, thank you for clarifying.

MOZGIII:
I noticed some tests failed - and the reason seems to be infrastructure issue, not the code issue.
The testing setup used here is not trivial, and I didn't add any new tests in this PR so far. Is it ok? If not - please help me with this.

test/ts/src/testRpcCombinations.ts
```diff
@@ -54,7 +55,8 @@ export function runWithHttp1AndHttp2(cb: (config: TestConfig) => void) {
 
 export function runWithSupportedTransports(cb: (transport: grpc.TransportConstructor | undefined) => void) {
   const transports: {[key: string]: grpc.TransportConstructor | undefined} = {
-    "defaultTransport": undefined
+    "defaultTransport": undefined,
+    "fetchWithOptions": fetchRequestWithOptions({}),
```
johanbrandhorst: I assume we may need to add this key to some test cases as well?

MOZGIII:
(I think) I added the new API to tests. Don't really know if it's the correct way - let's see what CI does in response to this.

test/ts/src/testRpcCombinations.ts
MOZGIII: This seems to be used by all test cases as a way to provide transport - so it sohlud do the trick. Testing setup is rather complictaed, but seems to be prepared for testing custom transports like this. Would like a comment from @MarcusLongmuir on this.

johanbrandhorst:
The tests probably won't pass until #262 is merged regardless, but we can force this through if everything but Chrome-42 passes.

MOZGIII:
What's going on with tests? They seems to be stuck - is it Travis CI that's not scheduling them to run?

johanbrandhorst:
I think it's just limited to 5 concurrent runners per project, and some of the runners are taken up by my #262. Please wait :).

MOZGIII:
From the previous test run: Safari now has some strange failures - https://travis-ci.org/improbable-eng/grpc-web/builds/443570174
Can you please take a look? Doesn't seem to be caused by my chages.

MOZGIII:
I uncovered another problem - transport implementations are not exported. I think I have to add a way to use this `fetchRequestWithOptions` thing from outside explicitly.

johanbrandhorst:
If Firefox should default to MozXHR AFAIK - if we've broken that we need to fix it. Are there tests running on Firefox with fetch that shouldn't be? I'm not very familiar with the tests either I'm afraid, but you should be able to run it locally I assume?

johanbrandhorst:
If you want, you can try reaching out to @MarcusLongmuir or @jonny-improbable or @easyCZ on the grpc-web slack (see README.md).

MOZGIII:
I have to step out now, hopefully they'll see this PR and comment before I get back.

jonny-improbable:
Hmm, I have some reservations with this approach. The main reason being that one of the goals of grpc-web is to provide an abstraction over the underlying browser technology; hence why we have both `fetch` and `xhr` based transports. 

The problem with this approach is that the caller is specifying the concrete transport to use; whilst `fetch` will work fine in Chrome, it would fail if executed in Firefox because streaming of response bodies is currently behind a feature flag (see [here](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API#Browser_compatibility)).

In the past I've considered adding a `credentials` field to the `TransportOptions` object and then implementing some logic in the `XhrTransport` to set the `withCredentials` prop accordingly; however given that we now support a WebSocket transport, such a `credentials` option  no longer applies to all transports. 

My initial thinking might be to create a clear distinction between the 2 types of transport we have: 'Http' based and 'Socket' based (where the HttpTransport is an abstraction over `fetch` and `xhr` based on the capabilities of the host browser). Once this has been achieved we can have different `TransportOptions` for each type of transport.

A further step could be to make the `DefaultTransportFactory` be injectable so the user can make any 'global' transport configuration without having to configure each rpc call site.

WDYT?

MOZGIII:
Well, the idea of this PR is to utilize the pre-existing ability to specify custom transport to fine-tune the underlying transport. The real goal here is to have the ability to fine-tune the transports (in particular - just the `fetch` transport). This brings me to the point: even if the transpord it chosen automatically, I'd like to have a way to file-tune it's internal to better fit my needs. I also like the transport autodetection, but the ability to force the usage of a specific transport and to control it would be a must have as well.
We were able to quickly test (and ditch, cause it didn't work as we expected) the setup with `credentials: 'include'` in our internal dev process using the custom `fetch` transport copied form tree. This would be much better if we could just do it directly, without doing those "smart tricks".
So, it's not just about letting setting just the `credentials` (though current PR just provides that) - it's about file-tuning the transport internal implementation.

What you're proposing is all good, but providing knobs (as an opt in - for people that need those, like me when I had that particular usecase) for transport is, I'd say more important. And, since knobs for different transports will be different, a unified interface to all of the possible tweakable parameters is not required - and adding knobs is not therefore blocked by splitting Http/Stream transports.

When interface for setting internal implementation details is ready - additional logic (like unified tweaking for unified `Http` tranport family) can be build ot not of it; and I'd be happy to see that, but even more I'd love the ability to tweak transports as close and flexible to native APIs as possible.

johanbrandhorst:
There's something to be said for carefully adding API surface though, and if we can accomplish the same thing with a longer term goal, shouldn't we try that first? If we only need to expose credentials and they can both be shared between Fetch and XHR, lets do that now.

MOZGIII:
Sure, why not. Especially that this PR just does that (for fetch), and we can implement simialr thing for XHR. We can also implement a "default" transport "withOptions", that would allow passing credentials, and setting them for `fetch`/`XHR`, and either ignoring or thoring an error if it the transport picked is not fetch not XHR. Does this plan seems reasonable?

jonnyreeves:
Enjoying a glass of wine, I'm having an idea. Let me see if I can make some
time to code it up at the weekend.

MOZGIII:
Superseded by #265

PR #265, opened 2018-10-20
Configurable Transports and Extraction of the Node HTTP Transport

### LINKED ISSUE #191: Using grpc-web-client with Angular (angular-cli)
Whenever I tried to serve/build project I get:

ERROR in ./node_modules/grpc-web-client/dist/transports/nodeHttp.js
Module not found: Error: Can't resolve 'http' in '<path_to_project>\node_modules\grpc-web-client\dist\transports'
ERROR in ./node_modules/grpc-web-client/dist/transports/nodeHttp.js
Module not found: Error: Can't resolve 'https' in '<path_to_project>\node_modules\grpc-web-client\dist\transports'

Anyone encountered it ever?

It seems that require('http') from nodeHttp.js can't hit NodeJS's http...

jonny-improbable:
- [x] Allow the user to specify the default transport
- [x] Allow the user to customise individual Transports, beyond the common `TransportOptions` interface
- [x] Expose the underlying Fetch and XHR based transports
- [x] Expose proposed set of configuration options in `FetchTransportInit`
- [x] Refactor `src/transport` folder structure / package layout
- [x] Remove and extract the Node-HTTP Transport into its own npm module
- [x] Test coverage for `setDefaultTransport`
- [x] Revisit `docs/transport.md`
----

## Change Overview
Following on from the conversation in #261 this PR allows the user to configure the behavior of a given  Transport, separate from the pre-existing `TransportOptions` interface which deals with passing request state (ie: host, url, callbacks, etc).

To facilitate this change, the `transport` option passed to the `unary`, `invoke` and `client` functions is now expected to meet the new `grpc.TransportFactory` interface, ie: a func which can accept zero or more args and will return a `grpc.Transport` instance. 

```javascript
grpc.unary(BookService.GetBook, {
    request: getBookRequest,
    host: host,
    transport: grpc.HttpTransport({ credentials: 'include' }),
    onEnd: ( /*...*/ )
});
```

Note this results in a breaking change for anyone currently using the Websocket Transport; instead of specifying:

```
transport: gprc.WebsocketTransportFactory
```

They would instead now call:

```
transport: grpc.WebsocketTransport()
```

This change also spurred me to to fix #191 and #141 by extracting the 'Node HTTP' transport out from the gprc-web-client package and into a new module: grpc-web-node-http-client, which will be published to npm separately. 

As one thing leads to another, extracting node-grpc-web-node-http-transport led to implement `grpc.setDefaultTransport() which allows the user to specify which `TransportFactory` should be .invoked if none is supplied in the options.

## Breaking Changes
- `transport` option passed to `unary`, `invoke` and `client` is now expected to by an instance of `Transport`, not a reference to a function which will return a Transport when invoked.
- `grpc-web-client` no longer auto-selects a suitable transport in a NodeJS environment; users of grpc-web-client in a NodeJS environment will need to depend upon `grpc-web-node-http-transport` and make use of `grpc.setDefaultTranpsport()`.

## What's Next?
- The whole npm linking stuff is disgusting and hard to maintain. I've played around with [Lerna](https://github.com/lerna/lerna) and I feel it would dramatically simplify things; however it does require us to move around a tonne of files, and I don't want to make this diff any bigger than it already is!
- I don't feel that the name of the `TransportOptions` interface conveys the intention of this object well and could be confusing to future developers. Might be wise to rename it, however this would be a breaking change now we've exported it!


MOZGIII:
Sorry, I accidentally posted the comment above mid-edit. It's now ready.

jonnyreeves:
Yep all makes sense. A, B and C all feel achievable inside the scope of this PR. Let's revisit D when those have been achieved.

Thanks for your perspective

ts/src/transports/Transport.ts
```diff
@@ -27,40 +26,42 @@ export interface TransportOptions {
...
 
-function detectTransport(): TransportConstructor {
+export function HttpTransport(init: HttpTransportInit): TransportFactory {
   if (detectFetchSupport()) {
-    return fetchRequest;
-  }
-
-  if (detectMozXHRSupport()) {
-    return mozXhrRequest;
+    return FetchReadableStreamTransport({ credentials: init.withCredentials ? 'include' : 'same-origin' })
```
johanbrandhorst: How do you configure this to `omit` now?

MOZGIII: This implementation allows explicitly using `transport: FetchReadableStreamTransport({ credentials: "omit" })` in the invocation/client intead of `transport: HttpTransport({ /* whatever */ })`.
`HttpTransport` can't support `omit` because it's not in the "greatest common interface", as `XMLHttpRequest` can only allow `include` or `same-origin` - https://xhr.spec.whatwg.org/#the-send()-method see *credentials mode* at (6) there.

MOZGIII:
So far it's all good. :+1: 
I'd suggest extracting `HttpTransport` into a separate file. It's probably a good idea to restructure transports code a bit - with a single transport per file, and probably moving the base `Transport` class to a upper level directory - so that only transport implementations are left at the transports dir.

MOZGIII:
Idea of what can be exposted for each transport:

- `XMLHttpRequest`:
  - `headers` - probably easier to expose as key-value pairs field rather than native `XMLHttpRequest` interface;
  - `withCredentials`
  I considered `timeout`, but it seems exposing it might distrupt normal gRPC execution. However I'd still expose it with a disclamer that it is to be used on one's own risk.

- `fetch`:
  - `headers` - can be exposed but custom merging is required
  - `mode`
  - `credentials`
  - `cache`
  - `redirect`
  - `referrer`
  - `referrerPolicy`
  - `integrity` - might not make sense, but I don't see a reason why it wouldn't work technically
  - `keepalive`

johanbrandhorst:
As for the proposed transport knobs; I don't think we should expose header configuration, this is already possible with gRPC metadata.

MOZGIII:
It seems like metadata values map directly to headers. This kind of makes headers-metadata relationship strongly coupled. This implemnetation detail makes it so it's indeed no need to allow setting headers, because setting metadata already does that.
After consulting https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md I think it's ok, and can confirm no explicit support for settings headers is nesessary.

jonny-improbable:
Agreed on not adding headers; I can look into adding the additional configuration options to Fetch and adding some new test coverage.

One thing worth explicitly calling out is that I have removed the 'NodeJS' transport from being automatically selected in the environment. Given issues such as #203 and #204 I would suggest that we extract the node-transport into it's own package (maybe: `improbable-eng/grpc-web-node-http-transport`)? Thoughts?

johanbrandhorst:
Agreed, it might break existing clients but I think node should be a separate concern.

MOZGIII:
@jonny-improbable I'd rather force-add them even if TypeScript doesn't support it - it's kind of providing the full implementation for the spec. TypeScript should not be a limiting factor for implementation.
It could be added later, but that's an extra step - and that's I don't like.

jonny-improbable:
@MOZGIII thanks for the gentle encouragement, figured out I was being stupid - all good now :)

johanbrandhorst:
Isn't the reason we shouldn't include them because the minimum browser we aim to support don't support these settings? The risk now is that users implement things tat aren't supported in the browser's we promise support for. We must at the very least expose this compromise to users.

jonny-improbable:
@johanbrandhorst The way I read @MOZGIII's request is that developers should be able to specify and configure a concrete / very browser specific transport if the wish to; this is the case with the `FetchReadableStreamTransport` which will basically only work in Chrome and later versions of Edge.  Using this transport, the developer assumes that their end users are only using these browsers.

For everyone else, they should use the `HttpTransport` which provides the Greatest Common Divisor (GCD) of XHR and FetchRedableStream based transports.

jonny-improbable:
More work on this PR - moved the code for the NodeHttpTransport over to https://github.com/improbable-eng/grpc-web-node-http-transport/tree/master

MOZGIII:
@johanbrandhorst in general, when it comes to building projects, it's the user of the library who's in control of browser compatibility requirement, not the library owner. Owners, therefore, should strive for providing more flexibility, than less.
This is not true, however whne it comes to "opinitionated" libraries/frameworks - there the flexibiliy is usually explicitly reduced to enforce some particular way of doing things.
gRPC implemenations are opinionated too in a way, but I guess they should only be opinionated when it comes to following the spec. Forcing programmers to use only a particular subset of options is not as good.

MOZGIII:
Actually, I just noticed that the spread operator is used: https://github.com/improbable-eng/grpc-web/blob/84e10bacd082016be2b29ea7d793fb95540c12b4/ts/src/transports/http/fetch.ts#L80-L86

This means any underkying values can be passed, which is good.

It also means that `FetchTransportInit` can be just:
```typescript
type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>
export type FetchTransportInit = Omit<RequestInit, "headers" | "method" | "body" | "signal">
```

test/ts/node-src/node.spec.ts
```diff
@@ -4,5 +4,7 @@ process.env.NODE_TLS_REJECT_UNAUTHORIZED = "0";
 process.env.DISABLE_CORS_TESTS = "true";
 // Disable the WebSocket-based tests as they don't apply for Node environments
 process.env.DISABLE_WEBSOCKET_TESTS = "true";
+// Although this will be set by Travis CI; it won't be set locally.
+process.env.BROWSER = "nodejs";
```
jonny-improbable: As the comment suggest, this is required when running the tests locally; this was not an issue before because the `node-http` transport was automatically detected by default; however now `testRpcCombinations.ts` sniffs for this environment variable and configures the `NodeHttpTransport` accordingly.

ts/package.json
```diff
@@ -34,7 +34,6 @@
   },
   "devDependencies": {
     "@types/google-protobuf": "^3.2.5",
-    "@types/node": "^10.7.1",
```
jonny-improbable: With the removal of the node-http transport from grpc-web-client - this is no longer required.

ts/src/transports/http/xhr.ts
```diff
@@ -80,12 +91,35 @@ class XHR implements Transport {
...
+    this.xhr.overrideMimeType("text/plain; charset=x-user-defined");
+  }
+
   cancel() {
     this.options.debug && debug("XHR.abort");
     this.xhr.abort();
   }
 }
 
+export class MozChunkedArrayBufferXHR extends XHR {
```
jonny-improbable: Removed a tonne of duplicated code between the xhr and mozXhr files by using a classic Template Design Pattern.

johanbrandhorst:
Amazing work Jonny!

ts/docs/client.md
```diff
@@ -60,7 +60,7 @@ Sending multiple messages and indicating that the client has finished sending ar
 
 Most browser networking methods do not allow control over the sending of the body of the request, meaning that sending a single request message forces the finishing of sending, limiting these transports to unary or server-streaming methods only.
 
-For transports that do allow control over the sending of the body (e.g. websockets - coming soon), the client can optionally indicate that it has finished sending. This is useful for client-streaming or bi-directional methods in which the server will send responses after receiving all client messages. Usage with unary methods is likely not necessary as server handlers will assume the client has finished sending after receiving the single expected message.
+For transports that do allow control over the sending of the body (e.g. websockets), the client can optionally indicate that it has finished sending. This is useful for client-streaming or bi-directional methods in which the server will send responses after receiving all client messages. Usage with unary methods is likely not necessary as server handlers will assume the client has finished sending after receiving the single expected message.
```
jonny-improbable: Unrelated change, but whilst I was here I wanted to fix this.

ts/docs/transport.md
```diff
@@ -6,15 +6,15 @@ A “transport” in this context is a wrapper of one of these methods of creati
...
 ## How does grpc-web-client pick a transport?
 
-You can specify the transport that you want to use for a specific invocation through the `library` property in the [`client`](client.md), [`invoke`](invoke.md) and [`unary`](unary.md) function options.
+You can specify the transport that you want to use for a specific invocation through the `transport` property in the [`client`](client.md), [`invoke`](invoke.md) and [`unary`](unary.md) function options.
 
-If a transport is not specified then a transport factory is used to determine the browser’s compatible transports. See [Available Transports](#available-transports)
+If a transport is not specified then a grpc-web-client will use the specified default transport which can be set via `grpc.setDefaultTransport`. By default an `HttpTransport` is used; this transport is configured to not send credentials (cookies) to cross-origin resources. 
 
-If none are found then an exception is thrown.
+will fall back to a DefaultHttpTransport which works across the vast majority of browsers with some limitations. See [Available Transports](#available-transports)
```
johanbrandhorst: This doesn't look right

ts/src/transports/Transport.ts
```diff
@@ -27,40 +29,11 @@ export interface TransportOptions {
...
-    return mozXhrRequest;
-  }
 
-  if (detectXHRSupport()) {
-    return xhrRequest;
-  }
 
-  if (detectNodeHTTPSupport()) {
-    return httpNodeRequest;
-  }
```
MOZGIII: When you fiew this file it's easy to notice a lot of empty lines - we should probably clean them up.
Have you considered using [prettier](https://github.com/prettier/prettier)?

ts/src/transports/http/http.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export function CrossBrowserHttpTransport(init: CrossBrowserHttpTransportInit): TransportFactory {
+  if (detectFetchSupport()) {
+    const fetchInit: FetchTransportInit = {
+      credentials: init.withCredentials ? "include" : "same-origin"
+    };
+    return FetchReadableStreamTransport(fetchInit);
+  }
+  return XhrTransport({ withCredentials: init.withCredentials });
+}
```
MOZGIII: No newline at EOF - consider adding .editorconfig to this repo

MOZGIII:
There are some minor code-style relating things. Other that that everything looks good, great job!

jonny-improbable:
Thanks @MOZGIII and @johanbrandhorst

MOZGIII:
Should we expect a version bump soon?

PR #270, opened 2018-10-28
Define and Export RpcOptions interface

jonny-improbable:
This allows us to import it in ts-protoc-gen and avoid us having to duplicate the definition.

Note the motivation for making this change is that I've just realised that 0.7.0 will break ts-protoc-gen as it expects the `transport` property to be typed as a `TransportConstructor`, not a `TransportFactory`.

### DIFF
--- a/ts/src/client.ts
+++ b/ts/src/client.ts
@@ -8,10 +8,13 @@ import {MethodDefinition} from "./service";
 import {frameRequest} from "./util";
 import {ProtobufMessage} from "./message";
 
-export interface ClientRpcOptions {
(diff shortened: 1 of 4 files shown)


MOZGIII:
Hey guys, it's me again!
Wouldn't it be good if ts-protoc-gen was self-contained? This could be easily achieved by making generated code generic.
Then, to maintain the ease of use, we could figure out a way to efficiently specialize the generic types with the exact type (for `Metadata` and `RpcOptions`/`ClientRpcOptions`) from the `grpc` package.

Issue #228, opened 2018-09-04
Use lower case header keys in in-body trailers as mandated by gRPC-Web spec.


ktr0731:
I found some concerns when I have tested a gRPC Web server related to trailers.

here are the response trailers.

```
Content-Type: application/grpc+proto
Grpc-Status: 0
```

1. Why aren't trailer names using lower-case? The gRPC Web spec says, "use lower-case header/trailer names".
2. Why is `Grpc-Status` at the end? In order to the ABNF of gRPC response, `Status` must be next to `Length-Prefixed-Message`.

ref.
https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md
https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md

johanbrandhorst:
This is a great spot @ktr0731, and at least the first suggestion is an obviously incorrect implementation of the spec. I think this is simply a matter of using the default Go header formatting without thinking about it (https://golang.org/pkg/net/textproto/#CanonicalMIMEHeaderKey).

The second issue is a bit more unclear. Do you mean to suggest that status should come before the content type trailer? The gRPC HTTP2 spec says that `Trilers` are _after_ the `Length-Prefixed-Message`. I don't see how this contradicts that?

ktr0731:
Thank you @johanbrandhorst! I want to fix the first matter if I can afford to do.

About the second issue, the spec says:

```
Response → (Response-Headers *Length-Prefixed-Message Trailers) / Trailers-Only
Response-Headers → HTTP-Status [Message-Encoding] [Message-Accept-Encoding] Content-Type *Custom-Metadata
Trailers → Status [Status-Message] *Custom-Metadata
```

When a response has some `Length-Prefixed-Message`, Response is represented as `Response → Response-Headers *Length-Prefixed-Message Trailers`.
`Response-Headers` is represented as `Response-Headers → HTTP-Status [Message-Encoding] [Message-Accept-Encoding] Content-Type *Custom-Metadata`. Therefore, `Content-Type` is a part of `Response-Headers`, not `Trailers`.

johanbrandhorst:
So, if I understand correctly, you are simply saying that the content type shouldn't be in the in-body trailers? Do you know what the behaviour of the Envoy proxy is in this case?

johanbrandhorst:
[...]

PR #271, opened 2018-10-29
Fix gRPC Web spec violation related to header/trailer names

ktr0731:
As the result of a disscussion of https://github.com/improbable-eng/grpc-web/issues/228, we realized that we must use lower header/trailer names.
The changes maps header/trailer names to lower at the end of a request.

### DIFF
--- a/go/grpcweb/grpc_web_response.go
+++ b/go/grpcweb/grpc_web_response.go
@@ -66,6 +66,7 @@ func (w *grpcWebResponse) copyJustHeadersToWrapped() {
 }
 
 func (w *grpcWebResponse) finishRequest(req *http.Request) {
+	w.mapHeaderKeysToLower()
 	if w.wroteHeaders || w.wroteBody {
 		w.copyTrailersToPayload()
 	} else {
@@ -116,7 +117,7 @@ func (w *grpcWebResponse) copyTrailersToPayload() {
 
 func (w *grpcWebResponse) extractTrailerHeaders() http.Header {
 	flushedHeaders := w.wrapped.Header()
-	trailerHeaders := make(http.Header)
+	trailerHeaders := make(header)
 	for k, vv := range w.headers {
 		// Skip the pre-annoucement of Trailer headers. Don't add them to the response headers.
 		if strings.ToLower(k) == "trailer" {
@@ -134,5 +135,22 @@ func (w *grpcWebResponse) extractTrailerHeaders() http.Header {
 			trailerHeaders.Add(k, v)
 		}
 	}
-	return trailerHeaders
+	return trailerHeaders.toHTTPHeader()
+}
+
+func (w *grpcWebResponse) mapHeaderKeysToLower() {
+	mapped := make(header)
+	for k, vv := range w.headers {
+		lowerKey := strings.ToLower(k)
+		// Use append directly instead of mapped.Add because mapped.Add calls strings.ToLower(k)
+		for _, v := range vv {
+			// If trailer headers, map to lower these keys.
+			if lowerKey == "trailer" {
+				mapped[lowerKey] = append(mapped[lowerKey], strings.ToLower(v))
+			} else {
+				mapped[lowerKey] = append(mapped[lowerKey], v)
+			}
+		}
+	}
+	w.headers = mapped.toHTTPHeader()
 }
--- a/go/grpcweb/grpc_web_response.go
+++ b/go/grpcweb/grpc_web_response.go
@@ -114,9 +114,9 @@ func (w *grpcWebResponse) copyTrailersToPayload() {
 	w.wrapped.(http.Flusher).Flush()
 }
 
-func (w *grpcWebResponse) extractTrailerHeaders() header {
+func (w *grpcWebResponse) extractTrailerHeaders() trailer {
 	flushedHeaders := w.wrapped.Header()
-	trailerHeaders := header{make(http.Header)}
+	trailerHeaders := trailer{make(http.Header)}
 	for k, vv := range w.headers {
 		// Skip the pre-annoucement of Trailer headers. Don't add them to the response headers.
 		if strings.ToLower(k) == "trailer" {
--- a/go/grpcweb/grpc_web_response.go
+++ b/go/grpcweb/grpc_web_response.go
@@ -146,14 +146,8 @@ func (w *grpcWebResponse) mapHeaderKeysToLower() {
 	mapped := make(header)
 	for k, vv := range w.headers {
 		lowerKey := strings.ToLower(k)
-		// Use append directly instead of mapped.Add because mapped.Add calls strings.ToLower(k)
 		for _, v := range vv {
-			// If trailer headers, map to lower these keys.
-			if lowerKey == "trailer" {
-				mapped[lowerKey] = append(mapped[lowerKey], strings.ToLower(v))
-			} else {
-				mapped[lowerKey] = append(mapped[lowerKey], v)
-			}
+			mapped.Add(k, v)
 		}
 	}
 	w.headers = mapped.toHTTPHeader()
--- a/go/grpcweb/grpc_web_response.go
+++ b/go/grpcweb/grpc_web_response.go
@@ -66,7 +66,10 @@ func (w *grpcWebResponse) copyJustHeadersToWrapped() {
 }
 
 func (w *grpcWebResponse) finishRequest(req *http.Request) {
-	w.mapHeaderKeysToLower()
+	// HTTP/2 always uses lower header/trailer keys.
+	if req.ProtoMajor == 1 {
+		w.mapHeaderKeysToLower()
+	}
 	if w.wroteHeaders || w.wroteBody {
 		w.copyTrailersToPayload()
 	} else {
--- a/go/grpcweb/grpc_web_response.go
+++ b/go/grpcweb/grpc_web_response.go
@@ -77,9 +77,8 @@ func (w *grpcWebResponse) copyTrailersAndHeadersToWrapped() {
 	w.wroteHeaders = true
 	wrappedHeader := w.wrapped.Header()
 	for k, vv := range w.headers {
-		lowerKey := strings.ToLower(k)
 		// Skip the pre-annoucement of Trailer headers. Don't add them to the response headers.
-		if lowerKey == "trailer" {
+		if strings.ToLower(k) == "trailer" {
 			continue
 		}
 		// Skip the Trailer prefix
@@ -87,7 +86,7 @@ func (w *grpcWebResponse) copyTrailersAndHeadersToWrapped() {
 			k = k[len(http2.TrailerPrefix):]
 		}
 		for _, v := range vv {
-			wrappedHeader.Add(lowerKey, v)
+			wrappedHeader.Add(k, v)
 		}
 	}
 	w.writeCorsExposedHeaders()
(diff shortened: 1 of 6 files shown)


go/grpcweb/grpc_web_response.go
```diff
@@ -66,6 +66,10 @@ func (w *grpcWebResponse) copyJustHeadersToWrapped() {
 }
 
 func (w *grpcWebResponse) finishRequest(req *http.Request) {
+	// HTTP/2 always uses lower header/trailer keys.
+	if req.ProtoMajor == 1 {
+		w.mapHeaderKeysToLower()
```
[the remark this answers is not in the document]
jonnyreeves: Nice approach, I always forget what I powerful abstraction `io.Writer` can be in go

go/grpcweb/grpc_web_response.go
```diff
@@ -134,5 +138,22 @@ func (w *grpcWebResponse) extractTrailerHeaders() http.Header {
...
+}
+
+func (w *grpcWebResponse) mapHeaderKeysToLower() {
+	mapped := make(header)
+	for k, vv := range w.headers {
+		lowerKey := strings.ToLower(k)
+		// Use append directly instead of mapped.Add because mapped.Add calls strings.ToLower(k)
+		for _, v := range vv {
+			// If trailer headers, map to lower these keys.
+			if lowerKey == "trailer" {
```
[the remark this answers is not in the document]
ktr0731: I think it is more clear! 
I'm sorry, I cannot understand what "call it `GrpcHeaders`" means?

go/grpcweb/grpc_web_response.go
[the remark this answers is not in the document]
ktr0731: @johanbrandhorst 
It is looks good. Just like this?
```suggestion
		var mappedHeader header
		w.headers.Write(mappedHeader)
		w.headers = mappedHeader.toHTTPHeader()
```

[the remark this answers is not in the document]
johanbrandhorst: No... I think we need a custom type (like you've made), make it a composition of `http.Header` and just define a `Write() ([]byte, error)` which is almost a copy of the one I linked, but with header names lower-cased before writing. I think you can remove most of your other changes and just do that.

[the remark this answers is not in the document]
ktr0731: Could you please an example?
I cannot understand how a custom type which implements `Write(w io.Writer) error` will be used...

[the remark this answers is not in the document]
johanbrandhorst: It's not specific to this part of the code, it's a general comment on the whole PR. Basically, remove everything, create a new type `type grpcTrailer struct { http.Header }` and implement `func (g GRPCTrailer) Write(w io.Writer) error`. Then in the `copyTrailersToPayload`, simply replace the `trailers.Write(trailerBuffer)` with `grpcTrailer(trailers).Write(trailerBuffer)`. That's all.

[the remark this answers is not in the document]
ktr0731: I got it. I didn't realize it is a general comment 😭 
Thank you!

go/grpcweb/trailer.go
```diff
@@ -0,0 +1,136 @@
...
+
+	// If trailer headers, map to lower these keys.
+	if lowerKey == "trailer" {
+		value = strings.ToLower(value)
+	}
+	t.Header[lowerKey] = append(t.Header[lowerKey], value)
+}
+
+// Write writes a header in wire format.
+func (t trailer) Write(w io.Writer) error {
```
ktr0731: I've copied these methods from `http.Header` as it is.

go/grpcweb/trailer.go
```diff
@@ -0,0 +1,136 @@
...
+	if !ok {
+		ws = stringWriter{w}
+	}
+	kvs, sorter := t.sortedKeyValues(exclude)
+	var formattedVals []string
+	for _, kv := range kvs {
+		for _, v := range kv.values {
+			v = headerNewlineToSpace.Replace(v)
+			v = textproto.TrimString(v)
+			for _, s := range []string{strings.ToLower(kv.key), ": ", v, "\r\n"} {
```
ktr0731: Only `strings.ToLower(kv.key)` is difference point.

ktr0731:
Thank you for your review! Could you take review again?

go/grpcweb/trailer.go
```diff
@@ -0,0 +1,99 @@
...
+
+func (t trailer) Add(key, value string) {
+	lowerKey := strings.ToLower(key)
+
+	// If trailer headers, map to lower these keys.
+	if lowerKey == "trailer" {
+		value = strings.ToLower(value)
+	}
+	t.Header[lowerKey] = append(t.Header[lowerKey], value)
+}
```
johanbrandhorst: I don't think we need this function either. Lets just rely on `http.Header.Add`.

johanbrandhorst: Ok, having looked at this, how about we _keep_ this method and just remove the `Write` method (relying on `http.Header.Write`? This code is such a complicated optimisation mess that we might be able to get away with just a custom `Add`.

ktr0731: Oh, just as you said, we can remove custom `Write` and rely on `http.Header.Write`!
I agree to remove custom `Write` and keep custom `Add` instead.

ktr0731: I've removed it.

go/grpcweb/trailer.go
```diff
@@ -0,0 +1,26 @@
...
+	http.Header
+}
+
+func (t trailer) Add(key, value string) {
+	lowerKey := strings.ToLower(key)
+
+	// If trailer headers, map to lower these keys.
+	if lowerKey == "trailer" {
+		value = strings.ToLower(value)
+	}
```
johanbrandhorst: Is this correct? The require is just for the header key to be lower case, right?

go/grpcweb/trailer.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+// gRPC-Web spec says that must use lower-case header/trailer names.
+// See "HTTP wire protocols" section in
+// https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md#protocol-differences-vs-grpc-over-http2
+type trailer struct {
+	http.Header
+}
+
+func (t trailer) Add(key, value string) {
```
jonnyreeves: @johanbrandhorst should we not also override the `Get` method and automatically lower-case the supplied `key` parameter?

I would be worried that developers may get confused if they were to `Add()` a header with an upper-case key and then not be able to `Get()` it at a later date - maybe I'm over thinking.

go/grpcweb/trailer.go
```diff
@@ -0,0 +1,136 @@
...
+	lowerKey := strings.ToLower(key)
+
+	// If trailer headers, map to lower these keys.
+	if lowerKey == "trailer" {
+		value = strings.ToLower(value)
+	}
+	t.Header[lowerKey] = append(t.Header[lowerKey], value)
```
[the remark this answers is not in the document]
johanbrandhorst: We really don't want to do that, the `http.Header.Add` method will upper case the key again.

go/grpcweb/trailer.go
johanbrandhorst: This is an unexported type, so it's not that important, and I don't think we use `trailer.Get` anywhere? I don't think it's necessary.

johanbrandhorst:
@ktr0731 just the last point about needing to lowercase values, then we can add tests and get this merged.

go/grpcweb/trailer.go
ktr0731: That's right. I remove it.

ktr0731:
Sorry for delay. I fixed the last point now.

ktr0731:
I'll write some tests from now.

go/grpcweb/wrapper_test.go
```diff
@@ -312,6 +311,14 @@ func (s *GrpcWebWrapperTestSuite) assertHeadersContainMetadata(headers http.Head
 	}
 }
 
+func (s *GrpcWebWrapperTestSuite) assertTrailersContainMetadata(trailers grpcweb.Trailer, meta metadata.MD) {
```
ktr0731: I copied `assertTrailersContainMetadata` and `assertHeadersGrpcCode` from `assertHeadersContainMetaData` and `assertTrailersGrpcCode` because `http.Header.Get` maps passed key to MIME header.

go/grpcweb/wrapper_test.go
```diff
@@ -340,14 +355,34 @@ func serializeProtoMessages(messages []proto.Message) [][]byte {
...
 	mimeHeader, err := tp.ReadMIMEHeader()
 	if err == nil {
 		return make(http.Header)
 	}
-	return http.Header(mimeHeader)
+
+	headers := make(http.Header)
+	bufferReader = bytes.NewBuffer(dataBytes)
+	tp = textproto.NewReader(bufio.NewReader(bufferReader))
+	for {
```
ktr0731: We should not use MIMEHeader to check whether response trailer names are lower-case because all trailer names will be normalized.
Instead of it, I extract trailer names from raw lines. Then replace MIME trailer names to it.

go/grpcweb/wrapper_test.go
```diff
@@ -340,14 +355,34 @@ func serializeProtoMessages(messages []proto.Message) [][]byte {
 	return out
 }
 
-func readHeadersFromBytes(dataBytes []byte) http.Header {
+func readHeadersFromBytes(t *testing.T, dataBytes []byte) http.Header {
```
johanbrandhorst: If this method is only used for reading trailers now (I can't see it used anywhere else), lets rename it `readTrailersFromBytes` and return `grpcweb.Trailer` instead?

go/grpcweb/wrapper_test.go
```diff
@@ -340,14 +355,34 @@ func serializeProtoMessages(messages []proto.Message) [][]byte {
...
+			break
+		}
+		require.NoError(t, err, "failed to read header line")
+
+		i := strings.IndexByte(line, ':')
+		if i == -1 {
+			require.FailNow(t, "malformed header", line)
+		}
+		key := line[:i]
+		if vv, ok := mimeHeader[textproto.CanonicalMIMEHeaderKey(key)]; ok {
```
johanbrandhorst: So we're using `mimeHeader` so that we don't have to parse the value from the line? Can we copy the logic from `ReadMimeHeader` and avoid the call to `ReadMimeHeader` altogether?

go/grpcweb/wrapper_test.go
ktr0731: Good idea, I'll rename it.

go/grpcweb/wrapper_test.go
ktr0731: Yes, we can copy from `ReadMIMEHeader`, but I've thought that logic of `ReadMIMEHeader` is complicated.
This is just test code, so we don't need to some optimization.
I think this is more clear than `ReadMIMEHeader` based logic.

johanbrandhorst: I see. Okay, can you just add a comment then? It took me a while to understand why we were reading the key but not the value.

johanbrandhorst:
Great work @ktr0731, lets just wait for tests to pass. I will update my grpc-web-compatibility-test repo as soon as we've merged this so we can test it again.

ktr0731:
Thank you for giving me your time to help.
I'm really pleasured to contribute to the great project!

johanbrandhorst:
Looks like everything passed, so I'll merge this :tada:

PR #273, opened 2018-11-02
Export a UMD Module

Dig-Doug:
Adds a [rollup](https://github.com/rollup/rollup) config that builds the library in multiple formats. Each build format is listed in the `package.json` so tools can leverage them.

The only noticeable change is that the output JS is now called `grpc-web-client.js` instead of `index.js`

Goes with this [js-browser-headers pull request](https://github.com/improbable-eng/js-browser-headers/pull/18).


johanbrandhorst:
Is this not going to be a breaking change? Renaming the main entry point I mean. I applaud this effort but we might want to avoid that, if it does. Also I know very little about this stuff so I'm not comfortable with reviewing it.

Dig-Doug:
I think it depends on how people are using the pre-built library:

1. If they're using a build system like webpack, webpack will use the [values in the package.json](https://webpack.js.org/configuration/resolve/#resolve-mainfields) to resolve which file to use. This change won't break them as the name is updated in the package.json

2. If they're manually copying the `dist/` files and/or making modifications, etc., this change will probably break them. 

There may be other usage patterns I don't know about (I'm by no means an expert), but overall I think renaming the exported file is a good idea so that it doesn't conflict with other libraries. Other popular libaries, e.g. [threejs](https://github.com/mrdoob/three.js/blob/dev/package.json#L5), [vuejs](https://github.com/vuejs/vue/blob/dev/package.json#L5), also use this pattern.

jonnyreeves:
Thanks for your contribution @Dig-doug. I'm guessing that the motivation behind this change is to allow consumers to access this library natively in a browser without having to use a common is module bundler.

If that's the case, then I agree this is a worthy goal, however I'm not convinced that the overhead of adding rollup (and the necessary typescript loader plugins) is justified especially given that none of the maintainers have experience with this tool chain.

I would be grateful if you could investigate if this could be achieved using the typescript compiler directly without the need to add rollup into the mix. However if others in the community feel this approach of beneficial then we can revisit this decision.

jonnyreeves:
If you can get this to work using webpack instead of rollup, that would be great. Thank you

Dig-Doug:
I've updated both PRs to use webpack. Webpack doesn't support ES module output right now, but I'm only interested in the UMD module so that's not an issue for me.

ts/package.json
```diff
@@ -2,11 +2,12 @@
   "name": "grpc-web-client",
   "version": "0.6.3",
   "description": "gRPC-Web client for browsers (JS/TS)",
-  "main": "dist/index.js",
-  "types": "dist/index.d.ts",
+  "main": "dist/grpc-web-client.js",
+  "browser": "dist/grpc-web-client.umd.js",
+  "types": "dist/typings/index.d.ts",
   "scripts": {
     "clean": "rm -rf build && rm -rf dist",
```
jonny-improbable: Nit: do we still need a `build/` folder? If not, let's remove it as part of this change.

jonny-improbable:
@Dig-Doug sorry to be a pain but I just landed #269 which has made some significant changes to the package structure.

To rebase your changes you will need to move all your changes aimed at `ts/*` to `client/grpc-web-client/*`.

Dig-Doug:
I tried rebasing -- it seems like all the files are correct, but the diff in GitHub is messed up (shows 106 files changed). Rather than try to fix this branch, I am going to open a new PR with a new branch based off head.

PR #276, opened 2018-11-06
Export a UMD Module

Dig-Doug:
Originally https://github.com/improbable-eng/grpc-web/pull/273, I thought it would be easier to start fresh instead of trying to fix a bad rebase.

### DIFF
--- a/client/grpc-web-client/package.json
+++ b/client/grpc-web-client/package.json
@@ -41,7 +41,6 @@
     "tslint": "^5.9.1",
     "typescript": "3.0.1",
     "ts-loader": "^2.0.1",
-    "tslib": "^1.9.3",
     "webpack": "^2.2.1"
   }
 }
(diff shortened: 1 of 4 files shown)


client/grpc-web-client/package.json
```diff
@@ -34,9 +35,13 @@
...
   },
   "devDependencies": {
+    "@types/node": "^7.0.5",
     "@types/google-protobuf": "^3.2.5",
     "google-protobuf": "^3.2.0",
     "tslint": "^5.9.1",
-    "typescript": "3.0.1"
+    "typescript": "3.0.1",
+    "ts-loader": "^2.0.1",
+    "tslib": "^1.9.3",
```
jonnyreeves: Is this dependency required?

Dig-Doug: No, removed.

PR #285, opened 2018-11-21
gRPC web client middleware [WIP]

easyCZ:
An implementation of the middleware pattern for the grpc-web-client.

* A middleware allows requests to be intercepted and side-effects performed, ie *logging* or *timing*
* Allows for a browser extension to be built acting as a middleware to show deserialized payloads, similar to Chrome's Network tab

Remaining TODO:
* [x] Tests
* [x] Example
* [x] Documentation

Due to the type change on the request properties, this should be assumed to be a non-backwards compatible change and will require bumping of a major (pre-1.0) version

### DIFF
--- a/client/grpc-web-client/src/client.ts
+++ b/client/grpc-web-client/src/client.ts
@@ -7,13 +7,23 @@ import {Transport, TransportFactory, makeDefaultTransport} from "./transports/Tr
 import {MethodDefinition} from "./service";
 import {frameRequest} from "./util";
 import {ProtobufMessage} from "./message";
+import {IdentityMiddleware, Middleware, MiddlewareConstructor} from "./middleware";
 
-export interface RpcOptions {
+export interface RpcOptions<
+  TRequest extends ProtobufMessage,
+  TResponse extends ProtobufMessage,
+  M extends MethodDefinition<TRequest, TResponse>,
+> {
   transport?: TransportFactory;
   debug?: boolean;
+  middleware?: MiddlewareConstructor<TRequest, TResponse, M>;
 }
 
-export interface ClientRpcOptions extends RpcOptions {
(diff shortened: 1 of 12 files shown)


client/grpc-web-client/src/client.ts
```diff
@@ -264,7 +298,8 @@ class GrpcClient<TRequest extends ProtobufMessage, TResponse extends ProtobufMes
     requestHeaders.set("content-type", "application/grpc-web+proto");
     requestHeaders.set("x-grpc-web", "1"); // Required for CORS handling
 
-    this.transport.start(requestHeaders);
+    const middlewareHeaders = this.middleware.onStart(requestHeaders) || requestHeaders;
```
jonnyreeves: IMHO it would be cleaner/easier to extend for the middleware.onStart interceptor to mutate the `Metadata` object instead of returning it; the `Metadata` object is, after-all fully mutable.

client/grpc-web-client/src/middleware.ts
```diff
@@ -0,0 +1,70 @@
+import {Metadata} from "./metadata";
+import {ProtobufMessage} from "./message";
+import {Code} from "./Code";
+import {MethodDefinition} from "./service";
+import {ClientRpcOptions} from "./client";
+
+const noop = () => void 0;
```
jonnyreeves: heh, for the 3 char saving, I think it's clearer to return `undefined` unless there's something I'm missing?

jonnyreeves:
This is looking really nice; I have one architectual observation around the `Metadata` object otherwise 👍 

I'm curious why you removed the debug middleware? This looked like a great use of middleware and would greatly simplify a lot of the code.

easyCZ:
Documentation and tests added. Tests are currently failing, will investigate later today.

PR #288, opened 2018-12-10
Provide hook to add values to context from websocket connection headers

kv1122:
I may be missing something, but I haven't been able to find another way to access the headers sent over the websocket connection in my service implementation on the server.

I've made a hook to access the headers after they are parsed by the wrapper and added values, in my case a session token, to the context.

```go
...
wrappedGrpc = grpcweb.WrapServer(grpcServer, grpcweb.WithWebsocketContextInterceptor(wsContextInterceptor))
...

func wsContextInterceptor(c ctx.Context, header http.Header) ctx.Context {
	token := header.Get(tokenHeaderKey)
	if token != "" {
		c = ctx.WithValue(c, tokenHeaderKey, token)
	}
	return c
}
```

### DIFF
--- a/go/grpcweb/options.go
+++ b/go/grpcweb/options.go
@@ -3,7 +3,10 @@
 
 package grpcweb
 
-import "net/http"
+import (
+	"context"
+	"net/http"
+)
 
 var (
 	defaultOptions = &options{
@@ -19,6 +22,7 @@ type options struct {
 	originFunc                     func(origin string) bool
 	enableWebsockets               bool
 	websocketOriginFunc            func(req *http.Request) bool
+	websocketContextInterceptor    func(ctx context.Context, header http.Header) context.Context
 }
 
 func evaluateOptions(opts []Option) *options {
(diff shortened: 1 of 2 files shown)


kv1122:
It's a bidirectional stream over websocket. I pass the token using the metadata on the client but it's not available on the raw request, it is, however, available after the wrapper parses the headers:

```go
...
headers, err := parseHeaders(string(readBytes))
...
```
Will investigate more. Would be nice to avoid this change:)

kv1122:
Like I say, I may be missing something, but I can't see a way to access the http.Request object, or the headers for that matter, in the service implementation. I can access it before it's passed to the wrapper, in the http handler method, but not in the service!

johanbrandhorst:
There's a good reason you can't access the raw HTTP information in the gRPC service - it's abstracted away. If you're doing some custom handling in the client where you're using raw http headers rather than gRPC metadata, then that won't be provided to the gRPC service. Is there a reason you can't use gRPC metadata?

kv1122:
Seems I was missing something:) 

How to access headers in my service implementation:

```go
md, ok := metadata.FromIncomingContext(ctx)
```
Thanks for your time Johan

Closing this now

johanbrandhorst:
Ah, I see why it was confusing to you. Glad you found a solution!

PR #292, opened 2018-12-22
grpcweb: Support grpc-web-text content-type

evanj:
This works using the official grpc-web client, at least for unary requests.
Websockets should explicitly fail, but I did not test them.
I don't use streaming, so that also is likely to not quite work correctly.


evanj:
This definitely "works". I haven't tried stream requests because I don't need them yet, so my interop test doesn't include support for them. For similar reasons, this will just return an error for web sockets. I'm happy to put a bit more work into this to "finish" it, but I probably need a bit of help at least to test those use cases.

johanbrandhorst:
@evanj this is great, thank you again for starting this. I think we'll definitely want some tests in but the proper way to test this would be with a client using it, of course. Maybe we could hardcode some requests and responses if we're not going to take the time to implement this functionality in the client?

Also, might be too late now, but I've created an interop repo for testing the various clients and proxies, might be useful for you to properly test streaming grpc-web-text: github.com/johanbrandhorst/grpc-web-compatibility-tests.

go/grpcweb/wrapper.go
```diff
@@ -105,10 +111,12 @@ func (w *WrappedGrpcServer) IsGrpcWebSocketRequest(req *http.Request) bool {
...
 // with the gRPC-Web protocol.
 func (w *WrappedGrpcServer) HandleGrpcWebRequest(resp http.ResponseWriter, req *http.Request) {
-	intReq := hackIntoNormalGrpcRequest(req)
-	intResp := newGrpcWebResponse(resp)
+
+	intReq, isTextFormat := hackIntoNormalGrpcRequest(req)
+	intResp := newGrpcWebResponse(resp, isTextFormat)
 	w.server.ServeHTTP(intResp, intReq)
 	intResp.finishRequest(req)
+	grpclog.Infof("WTF textFormat:%v finished", isTextFormat)
```
jonnyreeves: Left over debug statement?

go/grpcweb/wrapper.go
```diff
@@ -105,10 +111,12 @@ func (w *WrappedGrpcServer) IsGrpcWebSocketRequest(req *http.Request) bool {
 // layer to transform it to a standard gRPC request for the wrapped gRPC server and transforms the response to comply
 // with the gRPC-Web protocol.
 func (w *WrappedGrpcServer) HandleGrpcWebRequest(resp http.ResponseWriter, req *http.Request) {
-	intReq := hackIntoNormalGrpcRequest(req)
-	intResp := newGrpcWebResponse(resp)
+
```
jonnyreeves: Please remove empty line

go/grpcweb/wrapper.go
```diff
@@ -193,11 +207,49 @@ func (w *WrappedGrpcServer) isRequestForRegisteredEndpoint(req *http.Request) bo
...
-	return req
+	incomingContentType := grpcWebContentType
+	isTextFormat := strings.HasPrefix(contentType, grpcWebTextContentType)
+	if isTextFormat {
+		// body is base64-encoded: decode it
+		req.Body = decodeBase64Reader(req.Body)
+		incomingContentType = grpcWebTextContentType
+	}
+	req.Header.Set("content-type", strings.Replace(contentType, incomingContentType, grpcContentType, 1))
+	grpclog.Infof("WTF in:%s search:%s out:%s", contentType, incomingContentType, req.Header.Get("content-type"))
```
jonnyreeves: Left over debug statement?

jonnyreeves:
I should be able to have a go at modifying the client to support grpcwebtext which will give us an added layer of integration test coverage

evanj:
@jonnyreeves Added as a collaborator to my repo, and I removed the sloppy debugging stuff I left in there (oops!). Thanks for taking a look at this PR. I definitely appreciate the help getting this cleaned up and properly integrated.

I'm working on a Python gprc-web implementation, and mostly tested this to try to verify the interoperability to make sure I didn't screw stuff up :)

go/grpcweb/wrapper.go
```diff
@@ -193,11 +205,61 @@ func (w *WrappedGrpcServer) isRequestForRegisteredEndpoint(req *http.Request) bo
...
+
 	contentType := req.Header.Get("content-type")
-	req.Header.Set("content-type", strings.Replace(contentType, "application/grpc-web", "application/grpc", 1))
-	return req
+	incomingContentType := grpcWebContentType
+	isTextFormat := strings.HasPrefix(contentType, grpcWebTextContentType)
+	if isTextFormat {
+		// body is base64-encoded: decode it; Wrap it in readerCloser so Body is still closed
+		decoder := base64.NewDecoder(base64.StdEncoding, req.Body)
+		req.Body = &readerCloser{decoder, req.Body}
```
johanbrandhorst: Nit: can we make this explicitly assign the fields for clarity, I think this is a vet error now?

evanj: Done! Although this doesn't trigger any vet warnings, since this is a struct declared in this package. Declaring too many/few fields causes a compiler error with "too few values in &readerCloser literal", or if this was in a different package vet would warn about "composite literal uses unkeyed fields".

I also took this opportunity to tweak the initialization of `grpcWebResponse` in `newGrpcWebResponse` and remove an unused function from my previous io.Pipe related misadventure.

go/grpcweb/wrapper.go
```diff
@@ -158,13 +164,19 @@ func (w *WrappedGrpcServer) handleWebSocket(wsConn *websocket.Conn, req *http.Re
 	req.Method = http.MethodPost
 	req.Header = headers
 
-	w.server.ServeHTTP(respWriter, hackIntoNormalGrpcRequest(req.WithContext(ctx)))
+	// TODO: pass isTextFormat into newWebSocketResponseWriter and figure out how that works
```
johanbrandhorst: Can we add an issue to track the work on this and reference it here, if it's not going in at the same time?

go/grpcweb/grpc_web_response.go
```diff
@@ -18,10 +21,29 @@ type grpcWebResponse struct {
 	wroteBody    bool
 	headers      http.Header
 	wrapped      http.ResponseWriter
+
+	// Data in the body of the response must be written to writer.
+	writer io.Writer
```
johanbrandhorst: With the addition of this, what are we still using the `wrapped` field for? Can we shrink its interface definition from `http.ResponseWriter`?

client/grpc-web-client/src/client.ts
```diff
@@ -261,7 +262,13 @@ class GrpcClient<TRequest extends ProtobufMessage, TResponse extends ProtobufMes
...
+
+    if (this.props.format === "text") {
+      requestHeaders.set("content-type", "application/grpc-web-text");
+      requestHeaders.set("accept", "application/grpc-web-text");
+    } else {
+        requestHeaders.set("content-type", "application/grpc-web+proto");
+    }
     requestHeaders.set("x-grpc-web", "1"); // Required for CORS handling
 
     this.transport.start(requestHeaders);
```
johanbrandhorst: I assume the client work here is not completed? I don't see any base64 encoding/decoding of payloads.

jonnyreeves: No this is my WIP, tbh it's going a slightly bigger refactor than I first hoped as the transports need a fair bit of work - @johanbrandhorst shall we break that out into a separate PR and just have this one be to introduce the server-side changes?

johanbrandhorst: Well, it would be cool to get this in sooner rather than later so @evanj can get on with the work, if we're happy to wait with the automated tests until the frontend is done, I don't think that's a bad idea!

go/grpcweb/wrapper.go
```diff
@@ -22,6 +24,10 @@ var (
 	}
 )
 
+const grpcContentType = "application/grpc"
+const grpcWebContentType = "application/grpc-web"
+const grpcWebTextContentType = "application/grpc-web-text"
```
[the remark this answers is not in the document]
evanj: I've added a link to https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md#protocol-differences-vs-grpc-over-http2

These constants are actually used to search and replace the `Content-Type` header from the underlying gRPC server, so I think we do not want to include the `+whatever` suffixes. In theory: we could use use gRPC-Web with thrift or flatbuffers or whatever. I *think* this implementation will work, because it will just search/replace `application/grpc+flatbuffers` into `application/grpc-web-text+flatbuffers` and do the correct base64-encoding.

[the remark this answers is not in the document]
johanbrandhorst: I mention it because it seems we've hardcoded `+proto` in the client code.

go/grpcweb/wrapper.go
evanj: I created #294 to record this. I've removed this TODO.

go/grpcweb/grpc_web_response.go
johanbrandhorst: Hm, in my mind the Base64ResponseWriter would seem nice here, as opposed to two writers I think. Lets try that if you don't mind?

client/grpc-web-client/src/client.ts
johanbrandhorst: @evanj lets remove this from this PR to avoid blocking while @jonnyreeves looks at what would need to be done. I imagine we can ship this to master and when we have the functionality in the frontend we can release a minor version bump.

go/grpcweb/wrapper.go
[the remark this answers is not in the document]
evanj: Oh good point, I had not noticed that. I am *far* from knowledgeable enough about gRPC, gRPC-web, or this implementation to have a good answer here. I just know enough to have implemented a Python server, and hacked support into `grpcwebproxy` that seems to work :)

Now that I look at this briefly, I *think* this mismatch is fine: The Go server part is a proxy around the "official" gRPC server, which attempts to be at least somewhat protocol agnostic. The grpc-web client in this repository is opinionated, and implements a type of `export interface Client<TRequest extends ProtobufMessage, TResponse extends ProtobufMessage>`, so it will only ever be used with the `+proto` variant.

For what it is worth, this is [documented as being the default if missing](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md#protocol-differences-vs-grpc-over-http2). The "official" upstream client seems to omit the `+proto` string, in either the "native" node implementation (which seems to default to `application/grpc`), or the grpc-web Javascript (which defaults to `application/grpc-web-text`)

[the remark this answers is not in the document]
johanbrandhorst: Indeed, I agree with you, so let's leave it as it is :).

client/grpc-web-client/src/client.ts
evanj: Removed!

go/grpcweb/grpc_web_response.go
evanj: I gave this a shot. This ended up changing the output slightly, but in a way that is supported at least by the official grpc-web client (my personal "interop test" doesn't use the client in this repo).

evanj: Also I should note: thanks for pushing on this, I think the end result actually is cleaner.

go/grpcweb/wrapper_test.go
```diff
@@ -111,15 +132,48 @@ func (s *GrpcWebWrapperTestSuite) makeRequest(verb string, method string, header
 	return resp, err
 }
 
-func (s *GrpcWebWrapperTestSuite) makeGrpcRequest(method string, reqHeaders http.Header, requestMessages [][]byte) (headers http.Header, trailers grpcweb.Trailer, responseMessages [][]byte, err error) {
+func decodeMultipleBase64Chunks(b []byte) ([]byte, error) {
+	// grpc-web allows multiple base64 chunks: the implementation may send base64-encoded
+	// "chunks" with potential padding whenever the runtime needs to flush a byte buffer"
```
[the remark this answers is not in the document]
evanj: Removed.

go/grpcweb/grpc_web_response.go
```diff
@@ -136,3 +155,45 @@ func (w *grpcWebResponse) extractTrailerHeaders() trailer {
...
+
+func (w *base64ResponseWriter) WriteHeader(code int) {
+	w.wrapped.WriteHeader(code)
+}
+
+func (w *base64ResponseWriter) Flush() {
+	// Flush the base64 encoder by closing it; unfortunately we must ignore the error
+	// grpc-web allows the body to contain multiple padded base64-parts:
+	// https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md
+	w.encoder.Close()
```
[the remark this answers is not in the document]
evanj: Great idea.

evanj:
Thanks! I also added a unit test for `base64ResponseWriter` that I forgot to `git add` the last time.

johanbrandhorst:
Thanks again for this @evanj, I've confirmed this is working with a grpc/grpc-web client in text mode independently. I will update my pending grpc-web blog post and the [grpc-web-compatibility-tests](https://github.com/johanbrandhorst/grpc-web-compatibility-test) repo.

PR #295, opened 2019-01-10
Fix grpc-web client package name and link.

jonny-improbable:


johanbrandhorst:
Thanks for doing this Jonny, I will update the gRPC-Web blog on grpc.io

johanbrandhorst:
I just noticed btw, we probably want to update the npm badge link as well?

Issue #289, opened 2018-12-13
Migrate npm package name to @improbable-eng/grpc-web


jonnyreeves:
## Overview
- [Google's grpc-web implementation](https://github.com/grpc/grpc-web) is published to npm as [`grpc-web`](https://www.npmjs.com/package/grpc-web)
- [Improbable's gprc-web implementation](https://github.com/improbable-eng/grpc-web) (this project) is published to npm as [`grpc-web-client`](https://www.npmjs.com/package/grpc-web-client)

To disambiguate the two, I propose that we migrate this project to publish under the `@improbable-eng` namespace in npm, ie: `@improbable-eng/grpc-web`, making use the npm [orgs](https://docs.npmjs.com/orgs/) feature.

## Migration Path
- [x] Rename and publish 0.8.0 under `@improbable-eng/grpc-web` on npm (#290)
- [x] Update mono-repo to remove all references to `grpc-web-client` package (#297)
- [x] Release a new version of ts-protoc-gen which generates code that imports `@improbable-eng/grpc-web` (improbable-eng/ts-protoc-gen#147)
- [x] Publish grpc-web-node-http-transport under the @improbable-eng org. (#299)
- [x] Deprecate the current `grpc-web-client` and `grpc-web-node-http-transport` packages with a note that encourages consumers to switch to `@improbable-eng` versions

Consumers would need to modify their import statements accordingly:

```
# Before
import { grpc } from 'grpc-web-client'

# After
import { grpc } from '@improbable-eng/grpc-web'
```

cc @easyCZ @johanbrandhorst @MarcusLongmuir

jonny-improbable:
To get the ball rolling I've secured the `improbable-eng` org on npm

jonny-improbable:
Step one complete, [published 0.8.0 of @improbable-eng/grpc-web to npm](https://www.npmjs.com/package/@improbable-eng/grpc-web)

johanbrandhorst:
https://github.com/improbable-eng/ts-protoc-gen/blob/f1270468fc080e3b15e582892c589933d99af8f1/src/service/grpcweb.ts#L174

johanbrandhorst:
https://github.com/improbable-eng/ts-protoc-gen/pull/147

jonnyreeves:
ts-protoc-gen v0.9.0 has been released which code-gens import statements that use the `@improbable-eng/grpc-web` package.

jonny-improbable:
```
[...]

PR #297, opened 2019-01-12
Update all child packages to reference @improbable-eng/grpc-web

jonny-improbable:
Part of #289; this PR aims to remove all references to the `grpc-web-client` package from the monorepo.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,6 @@
 {
   "name": "grpc-web-ci",
+  "private": true,
   "scripts": {
     "postinstall": "npm run bootstrap",
     "bootstrap": "lerna bootstrap",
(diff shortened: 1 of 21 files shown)


integration_test/ts/_proto/improbable/grpcweb/test/test_pb.js
```diff
@@ -251,7 +251,7 @@ proto.improbable.grpcweb.test.PingRequest.prototype.getValue = function() {
 
 /** @param {string} value */
 proto.improbable.grpcweb.test.PingRequest.prototype.setValue = function(value) {
-  jspb.Message.setField(this, 1, value);
+  jspb.Message.setProto3StringField(this, 1, value);
```
jonny-improbable: Hmm no sure why these changes have crept in, let me check my local env.

package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "grpc-web-ci",
+  "private": true,
```
jonny-improbable: Unrelated, but the root package should never be published as it just contains metadata for the monorepo.

PR #299, opened 2019-01-13
Migrate grpc-web-node-http-transport package so it's under the @improbable-eng org. on npm.

jonny-improbable:
To further disambiguate that this transport is specifically for use with the `@improbable-eng/grpc-web` package.

Note I've also fixed an issue with the dependencies; `@improbable-eng/grpc-web` should have been listed as a peer-dependency as it's intended to be used as a plugin (ie: the consumer is expected to have `@improbable-eng/grpc-web` somewhere in their dependency tree).

### DIFF
--- a/client/grpc-web-node-http-transport/package.json
+++ b/client/grpc-web-node-http-transport/package.json
@@ -1,6 +1,6 @@
 {
-  "name": "grpc-web-node-http-transport",
-  "version": "0.0.1",
+  "name": "@improbable-eng/grpc-web-node-http-transport",
+  "version": "0.0.2",
   "description": "Node HTTP Transport for use with @improbable-eng/grpc-web",
   "main": "lib/index.js",
   "repository": {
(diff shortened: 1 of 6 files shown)


client/grpc-web/package.json
```diff
@@ -27,7 +27,6 @@
     "dist"
   ],
   "peerDependencies": {
-    "@types/google-protobuf": "^3.2.5",
```
jonny-improbable: This package provides TypeScript definitions which are only required during compilation (and already satisfied as a devDependency); consumers do not need to fulfil this dependency as they are not expected to compile this package.

client/grpc-web-node-http-transport/package.json
```diff
@@ -13,13 +13,15 @@
...
-  },
   "author": "",
   "license": "Apache-2.0",
+  "peerDependencies": {
+    "@improbable-eng/grpc-web": ">=0.7.0"
+  },
   "devDependencies": {
+    "@improbable-eng/grpc-web": "^0.8.0",
     "@types/node": "^10.12.0",
+    "google-protobuf": "^3.2.0",
```
jonny-improbable: ...to satisfy `@improbable-eng/grpc-web`'s peerDependency

Issue #325, opened 2019-02-05
Eliminate custom header x-grpc-web


sgammon:
As far as I can tell, there is no actual functional purpose for the header `x-grpc-web`, aside from a vanity bump that it is possible to detect use of Improbable's gRPC web client.

Why is the header added? Of course there are significant problems with this approach. If gRPC web is being used cross-domain, the `x-grpc-web` header _must_ be included in the *Access-Control-Allow-Headers* response. And, naturally, the custom header triggers a CORS pre-flight which costs an entire HTTP round-trip, and is a head-of-line blocker for initial calls to the server.

So, here is my question:
1) Why was this header added in the first place?
2) Perhaps related to point 1, does this header _actually_ trigger any functionality? Is it triggered when using Improbable's server side tools, or all gRPC servers?
3) Why can't it be disabled, since the library works fine if the header is eliminated?

johanbrandhorst:
I can't answer for #1, but I think I can answer the others;

2. No, it does not trigger any functionality in the proxy, the traffic is split based on the content-type header.
3. I think it could probably be disabled. It's not required by the spec (https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md), though there's another header we might need to worry about instead: X-User-Agent.

My best guess is that the header is leftover from the olden days of before the spec. @marcuslongmuir probably knows.

johanbrandhorst:
I suppose removing this header might constitute a breaking change, so I've labeled it as such. We're still in 0.x.y so it's not the end of the world.

sgammon:
Thank you @johanbrandhorst for the quick response. On our end we manually remove the line in question before using the library. So we are well equipped to test if you'd like feedback.

Otherwise we will happily upgrade when this is merged.

MarcusLongmuir:
This was originally added for our own internal usage and then not removed. As far as I can tell the only thing that would break if you removed the client's sending of this header would be the [CORS handling](https://github.com/improbable-eng/grpc-web/blob/master/go/grpcweb/wrapper.go#L188). 

Unfortunately that would mean that anyone using this package that upgraded their client before their server would be affected, unless we remove the server usage first.

Alternatively, if you just need the client to not send this we can put in an option (e.g. `sendLegacyHeaders` which defaults to `true` and sends the header.) and then later remove this option (and stop sending the header).

johanbrandhorst:
[...]

PR #158, opened 2018-04-01
Deny CORS Requests by Default

jonny-improbable:
Change the default implementation of `grpcweb.WrapServer` so it denies requests from foreign origins by default to avoid circumventing the browser security model.

See discussion in #61.

PR #331, opened 2019-02-06
Fix: Eliminate x-grpc-web header

sgammon:
This changeset removes the custom `x-grpc-web` header, in favor of protocol-compliant content type sniffing on the server side. Fixes and closes improbable-eng/grpc-web#325.

So far, changes made:
- Modify Go server code to sniff for content type beginning with `application/grpc`
- Remove custom header from client code, so it is not sent

Curious for feedback from @johanbrandhost and @MarcusLongmuir. We don't use the Go server in production, we use Endpoints Service Proxy and Envoy.

### DIFF
--- a/client/grpc-web/src/client.ts
+++ b/client/grpc-web/src/client.ts
@@ -262,7 +262,6 @@ class GrpcClient<TRequest extends ProtobufMessage, TResponse extends ProtobufMes
 
     const requestHeaders = new Metadata(metadata ? metadata : {});
     requestHeaders.set("content-type", "application/grpc-web+proto");
-    requestHeaders.set("x-grpc-web", "1"); // Required for CORS handling
 
     this.transport.start(requestHeaders);
   }
--- a/go/grpcweb/wrapper.go
+++ b/go/grpcweb/wrapper.go
@@ -184,8 +184,7 @@ func (w *WrappedGrpcServer) IsGrpcWebRequest(req *http.Request) bool {
 //
 // You can control the CORS behaviour using `With*` options in the WrapServer function.
 func (w *WrappedGrpcServer) IsAcceptableGrpcCorsRequest(req *http.Request) bool {
-	accessControlHeaders := strings.ToLower(req.Header.Get("Access-Control-Request-Headers"))
-	if req.Method == http.MethodOptions && strings.Contains(accessControlHeaders, "x-grpc-web") {
+	if req.Method == http.MethodOptions && strings.HasPrefix(req.Header.Get("content-type"), grpcContentType) {
 		if w.opts.corsForRegisteredEndpointsOnly {
 			return w.isRequestForRegisteredEndpoint(req)
 		}


client/grpc-web/src/client.ts
```diff
@@ -262,7 +262,6 @@ class GrpcClient<TRequest extends ProtobufMessage, TResponse extends ProtobufMes
 
     const requestHeaders = new Metadata(metadata ? metadata : {});
     requestHeaders.set("content-type", "application/grpc-web+proto");
-    requestHeaders.set("x-grpc-web", "1"); // Required for CORS handling
```
sgammon: remove header (note that `content-type` begins with `application/grpc`)

go/grpcweb/wrapper.go
```diff
@@ -184,8 +184,7 @@ func (w *WrappedGrpcServer) IsGrpcWebRequest(req *http.Request) bool {
 //
 // You can control the CORS behaviour using `With*` options in the WrapServer function.
 func (w *WrappedGrpcServer) IsAcceptableGrpcCorsRequest(req *http.Request) bool {
-	accessControlHeaders := strings.ToLower(req.Header.Get("Access-Control-Request-Headers"))
-	if req.Method == http.MethodOptions && strings.Contains(accessControlHeaders, "x-grpc-web") {
+	if req.Method == http.MethodOptions && strings.HasPrefix(req.Header.Get("content-type"), grpcContentType) {
```
sgammon: sniff `content-type` header using identical code from earlier in this same file

sgammon:
any feedback about these failing tests would be appreciated. things seem to work when `BROWSER=node` but not really anywhere else.

this seems to be a frequent culprit:
```
            Chrome 57.0.2987 (Windows 7 0.0.0)
          Uncaught AssertionError: expected [] to deeply equal [ '12' ] thrown
          ✖ should handle a streaming response of multiple messages
```

i presume this could be a matching error based on the header that is no longer there, or perhaps it could be masking something more sinister (i.e. a CORS failure due to the headers change).

if someone knows what i can do to fix, i'm happy to apply it, push, and re-run CI

johanbrandhorst:
I'd like to see this change go in, but I have no idea about what might be causing the test failures I'm afraid. I think @jonnyreeves, @MarcusLongmuir or @easyCZ might be best to help. Sorry for the delay, they're usually quite busy :). Might be worth digging into the code yourself if you need this quickly.

jonnyreeves:
I'll flag this up with @MarcusLongmuir on Monday as I think he will likely have the most context; it's worth noting that whilst our integration tests _are_ flakey, there's clearly something broke if all of them are consistently failing.

go/grpcweb/wrapper.go
jonny-improbable: Hey @sgammon, I just dug into this code with @MarcusLongmuir. 

It's important to understand that the client has no control over the `Options` request (it is made automatically by the browser) - this is known as the preflight request.

It's also important to note that the `content-type` header of the Options request is not the same as the `content-type` of the request that triggered the preflight request.

We believe the correct remedy is to:

1. Change this conditional check to simply look for Options requests, ie:

```
if req.Method == http.MethodOptions {
``` 

2. Fast follow on #158 (preferably in the same release as this change).

Give this a shot and hopefully the browser tests will pass :)

jonny-improbable:
@sgammon, commented inline [here](https://github.com/improbable-eng/grpc-web/pull/331#discussion_r255485593).

go/grpcweb/wrapper.go
sgammon: thank you @jonny-improbable, i am aware of the distinction between the pre-flight and the `content-type` header here, but it's not my code, so i figured i would chip in to fix it because it breaks so frequently without custom CORS config

that being said, i was *not* aware the `content-type` isn't carried over in the pre-flight, which does make sense. of course there should be ways to avoid a pre-flight entirely, namely, omitting a custom `content-type` entirely (which breaks protocol spec but avoids an initial round-trip) or even leveraging browser form submission, but yes, we are on the same page.

in our environment unless i am mistaken we:
1) omit the `x-grpc-web` header to avoid the `Access-Control-Allow-Headers` change it requires
2) omit the `content-type`, affix it server-side before dispatching via gRPC

between those two i *believe* we avoid the pre-flight but i'll check to make sure. perhaps it might be a good option to add to improbable's branch of grpc-web since it seems you guys have control of the entire stack and could make a decision against spec like that, if you wanted. but hiding it behind a flag would be good :)

i'll go ahead and apply the conditional change, i am completely unfamiliar with golang but i will take your word for it

jonny-improbable:
Moving the conversation out of [this comment thread](https://github.com/improbable-eng/grpc-web/pull/331#discussion_r256068737) for better visibility.

> perhaps it might be a good option to add to improbable's branch of grpc-web since it seems you guys have control of the entire stack and could make a decision against spec like that, if you wanted. but hiding it behind a flag would be good :)

As a whole, we are trying to get closer to the gprc-web spec rather than diverge from it further.

Having dug into this in more detail (see this comment) I believe the correct course of action is to replace the `x-grpc-web` header with the custom `X-User-Agent` header set out in the [grpcweb spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md) - note this would still result in a preflight. Given this, I feel it would be best to not merge this PR as is.

@sgammon if you feel there's a case for eliminating the preflight entirely then can I suggest you raise it with the grpcweb project (the suggestion outlined above sounds reasonable to me) and if you get traction we would be happy to continue to follow any future revisions of the spec this end.

Thank you for your efforts to help us untangle this.

sgammon:
@jonny-improbable i was in the middle of implementing your suggested changes - the comment there about eliminating the pre-flight is just an idea

AFAIK the `x-grpc-web` header, which is what this PR is about, is non-spec. right? or am i missing something?

if that's the case, why doesn't this PR work? in theory it would move it closer to spec.

sgammon:
as a user of the software, we only wish to be able to use gRPC web on its own, not with the gRPC go tools provided by improbable.

it works great for our needs but we find ourselves needing to eliminate the `x-grpc-web` header by hand, which is why this change was proposed.

should i fork, or should i continue implementing your suggestion from above?

johanbrandhorst:
Reopening following discussion in #325

johanbrandhorst:
@sgammon do you intend to pick this up?

sgammon:
@johanbrandhorst sorry, i just have no bandwidth for this. i'd be happy to test if someone else wants to finish up.

PR #333, opened 2019-02-06
Support client certs for backend connections

### LINKED ISSUE #332: Support client certs for backend connections
Add support for using client certificates for connections to the backend gRPC server

mastersingh24:
Added 3 flags to grpcwebproxy to support gRPC
backend which requires TLS client auth.

Fixes #332 

Signed-off-by: mastersingh24 <gari.r.singh@gmail.com>

### DIFF
--- a/go/grpcwebproxy/backend.go
+++ b/go/grpcwebproxy/backend.go
@@ -30,12 +30,6 @@ var (
 		"Whether to ignore TLS verification checks (cert validity, hostname). *DO NOT USE IN PRODUCTION*.",
 	)
 
-	flagBackendTlsClientCertRequired = pflag.Bool(
-		"backend_client_tls",
-		false,
-		"Whether the gRPC server of the backend requires client certificates.",
-	)
-
 	flagBackendTlsClientCert = pflag.String(
 		"backend_client_tls_cert_file",
 		"",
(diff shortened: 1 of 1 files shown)


go/grpcwebproxy/backend.go
```diff
@@ -30,6 +30,22 @@ var (
 		"Whether to ignore TLS verification checks (cert validity, hostname). *DO NOT USE IN PRODUCTION*.",
 	)
 
+	flagBackendTlsClientCertRequired = pflag.Bool(
+		"backend_client_tls",
+		false,
+		"Whether the gRPC server of the backend requires client certificates.",
+	)
```
johanbrandhorst: I don't think we need this flag; we can unconditionally add the client cert if it's provided.

go/grpcwebproxy/backend.go
```diff
@@ -103,5 +119,19 @@ func buildBackendTlsOrFail() *tls.Config {
 			}
 		}
 	}
+	if *flagBackendTlsClientCertRequired {
+		if *flagBackendTlsClientCert == "" {
```
johanbrandhorst: Change this check to just check the other two are set.

go/grpcwebproxy/backend.go
mastersingh24: ok ... I almost went that route originally

go/grpcwebproxy/backend.go
mastersingh24: Added a bit of logic to make sure they are both set and removed the flag

johanbrandhorst:
LGTM let's wait for the build to finish

johanbrandhorst:
I've restarted the failing tests, this change shouldn't be causing any test failures :thinking:

johanbrandhorst:
Round 3, GO!

johanbrandhorst:
Thanks for your contribution!

mastersingh24:
Cool.  Thanks.  And thanks for this project overall ... very cool.

Issue #334, opened 2019-02-07
grpc.WrappedGrpcServer.ServerHTTP() returns response with invalid Content-Type header


danilvpetrov:
When wrapping `grpc.Server` in type `grpc.WrappedGrpcServer` using package `github.com/improbable-eng/grpc-web/go/grpcweb`, the method `ServeHTTP` does not set a `Content-Type` header either to `application/grpc-web-text` or `application/grpc-web`. This causes a javascript client using protobuf generated files with `https://github.com/grpc/grpc-web` package to skip processing the response from the gRPC server. As a result, a callback function on the request method or on stream `data` event are not getting called even though the entire request is completed successfully (in case of unary request) or ongoing (in case of server-side streaming).

This behaviour of `grpc.WrappedGrpcServer` is manifested both when using a unary request and server-side streaming. Also the result is the same regardless of the mode the javascript files are generated with (`grpcwebtext` or `grpcweb`).

I am using the latest `https://github.com/grpc/grpc-web` library (1.0.3) and the [latest](https://github.com/improbable-eng/grpc-web/commit/412759a1acbb78a180199edd11898224528ecd4e) version of `github.com/improbable-eng/grpc-web/go/grpcweb` package

johanbrandhorst:
This seems like a regression to me, last time I ran my interop tests all the requests from a grpc/grpc-web client worked with the `grpcwebproxy` proxy. I'll try and reproduce this in https://github.com/johanbrandhorst/grpc-web-compatibility-test.

danilvpetrov:
Thanks for the prompt reply! 

https://github.com/johanbrandhorst/grpc-web-compatibility-test repository has slightly different set up as it uses `https://github.com/improbable-eng/grpc-web/tree/master/go/grpcwebproxy` as a compiled binary run within a Docker container. This behaviour most likely will not be revealed as gRPC-native servers will supply the correct `Content-Type` header when connected to the proxy. These headers then that will be correctly interpreted by `grpc.WrappedGrpcServer`.

[...]

PR #339, opened 2019-02-09
WIP: go/grpcweb: add failing test

johanbrandhorst:
Hopefully first step towards fixing #334.

The test `TestPingError_WithTrailersInHeaders` is failing to translate the content-type from `application/grpc` to `application/grpc-web`. With a failing test iterating on solving it should be easier. I'm not _entirely_ sure the existing tests cover all the cases when the content type isn't being translated properly but after fixing the test we can hopefully manually test all the other failing cases to see if it fixes it.

@danilvpetrov


johanbrandhorst:
Abandoning in favor of #359

PR #341, opened 2019-02-11
Add travis_retry

jonny-improbable:
*should* retry each failed matrix build up to 3 times...


jonny-improbable:
Well they did retry (but still failed :D)

PR #358, opened 2019-02-20
Add Release Script

jonny-improbable:
- Triggered by travis when a new tag matching v*X.y.z* is pushed.
- Generates grpcwebproxy binaries for win, mac, and linux
- Uses `github-release` binary to do what it says on the tin and attach the grpcwebproxy binaries to the release
- Uses `lerna publish` to publish the client packages to npm

### TODO
- [ ] Verify that this works locally by publishing v0.9.0 to NPM and github
- [ ] Inject credentials into travis for creating github releases

### DIFF
--- a/go/generate-grpcwebproxy-binaries.sh
+++ b/go/generate-grpcwebproxy-binaries.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+rm -rf build
+mkdir -p build
+
+GOOS=linux GOARCH=amd64 go build -o build/grpcwebproxy-v0.8.1-linux-amd64 ./grpcwebproxy
+GOOS=windows GOARCH=amd64 go build -o build/grpcwebproxy-v0.8.1-windows-amd64.exe ./grpcwebproxy
+GOOS=darwin GOARCH=amd64 go build -o build/grpcwebproxy-v0.8.1-darwin-amd64 ./grpcwebproxy
--- a/go/generate-grpcwebproxy-binaries.sh
+++ b/go/generate-grpcwebproxy-binaries.sh
@@ -1,8 +1,0 @@
-#!/bin/bash
-
-rm -rf build
-mkdir -p build
-
-GOOS=linux GOARCH=amd64 go build -o build/grpcwebproxy-v0.8.1-linux-amd64 ./grpcwebproxy
-GOOS=windows GOARCH=amd64 go build -o build/grpcwebproxy-v0.8.1-windows-amd64.exe ./grpcwebproxy
-GOOS=darwin GOARCH=amd64 go build -o build/grpcwebproxy-v0.8.1-darwin-amd64 ./grpcwebproxy
--- a/publish-release.sh
+++ b/publish-release.sh
@@ -19,8 +19,13 @@ fi
 
 echo "Generating grpcwebproxy binaries"
 cd go
-./generate-grpcwebproxy-binaries.sh
-ls -l ./build
+rm -rf dist
+mkdir -p dist
+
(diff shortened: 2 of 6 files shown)


go/generate-grpcwebproxy-binaries.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+rm -rf build
+mkdir -p build
+
+GOOS=linux GOARCH=amd64 go build -o build/grpcwebproxy-v0.8.1-linux-amd64 ./grpcwebproxy
```
johanbrandhorst: Probably easier to inline this script in the release so you can use the tag name without passing bash arguments around.

johanbrandhorst:
As an alternative to this, have you looked at configuring some GitHub actions? We can use actions for both of these steps I think.

go/generate-grpcwebproxy-binaries.sh
jonny-improbable: Inline it into `publish-release.sh`? Ah actually yes, I see I'm hard coding the version string - doh!

publish-release.sh
```diff
@@ -0,0 +1,33 @@
...
+echo "Generating grpcwebproxy binaries"
+cd go
+./generate-grpcwebproxy-binaries.sh
+ls -l ./build
+cd ..
+
+echo "Publishing $TAG"
+
+# Create github release and attach server binaries
+./node_modules/.bin/github-release upload \
```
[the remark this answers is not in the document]
jonny-improbable: Are you referring to `github-release`? I've added it to the project's list of npm dependencies.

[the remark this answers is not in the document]
johanbrandhorst: Yeah, I just don't understand how travis works, I don't think we're checking in `./node_modules/`, does travis guarantee that it will run `npm install` before any script or something?

johanbrandhorst:
To add to my previous comment about github actions: https://sosedoff.com/2019/02/12/go-github-actions.html.

publish-release.sh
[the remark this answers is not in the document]
jonny-improbable: I believe so, however given our current flakey tests I am concerned that the publish step will never be reached as the test test that proceeds it will likely fail. 

You mentioned using Github Actions as an alternative - these are new to me, let me take a look.

publish-release.sh
```diff
@@ -0,0 +1,43 @@
...
+  echo "Invalid release tag: $TAG"
+  exit 1
+fi
+
+echo "Generating grpcwebproxy binaries"
+cd go
+rm -rf dist
+mkdir -p dist
+
+GOOS=linux GOARCH=amd64 go build -o "dist/grpcwebproxy-$TAG-linux-amd64" ./grpcwebproxy
```
johanbrandhorst: Sorry I didn't mention this before, but I think this is usually called `x86_64` in this context (ref: protobuf: https://github.com/protocolbuffers/protobuf/releases).

jonny-improbable: Thanks, follow conventions from protobuf repo and zipped all the binaries as well; updated the bins attached to the 0.9.0 release

johanbrandhorst:
Would be nice to have 0.9.1 for #359 btw.

johanbrandhorst:
bump

jonny-improbable:
Thanks for the bump @johanbrandhorst; just checked in with our eng-velocity team who are happy for me to go ahead and investigate github actions. I've taken the first step and signed up for the beta.

jonny-improbable:
Took the executive decision to merge this as-is.

Github never got back to me with with regards to adding us to the Github Hooks Beta, but seeing as I need to cut a release today, I'm going to invoke this script on my machine to at-least automate most of the process.

PR #359, opened 2019-02-20
Ensure headers from wrapped server is forwarded appropriately

### LINKED ISSUE #334: grpc.WrappedGrpcServer.ServerHTTP() returns response with invalid Content-Type header

danilvpetrov:
Fix #334.

This approach adds 'copyHeader' function with multiple options to address repeatable code in `go/grpcweb/grpc_web_response.go` file and header preparation when underlying gRPC service calls `Write` method on `grpcWebResponse` object.

The tests were copied from #339  originally done by @johanbrandhorst.

### DIFF
--- a/go/grpcweb/grpc_web_response.go
+++ b/go/grpcweb/grpc_web_response.go
@@ -45,13 +45,15 @@ func (w *grpcWebResponse) Header() http.Header {
 }
 
 func (w *grpcWebResponse) Write(b []byte) (int, error) {
-	w.wroteBody = true
+	if !w.wroteHeaders {
+		w.prepareHeaders()
+	}
+	w.wroteBody, w.wroteHeaders = true, true
 	return w.wrapped.Write(b)
 }
 
 func (w *grpcWebResponse) WriteHeader(code int) {
-	w.copyJustHeadersToWrapped()
-	w.writeCorsExposedHeaders()
+	w.prepareHeaders()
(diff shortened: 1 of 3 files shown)


johanbrandhorst:
Rerunning the tests :thinking:

danilvpetrov:
@johanbrandhorst: thanks for all your help!

PR #360, opened 2019-02-20
update example

atecce:
i'm trying to run the example and i think it's a _tad_ out of date. the only thing i've verified so far is this actually runs and listens on `:9091`, but i figured i'd just put it out there early and maybe update this as i play with it more.


johanbrandhorst:
Thanks a lot for this @atecce it does indeed look like we could do with these fixes. Assigning to @jonny-improbable as he's a bit more of a TS wiz than I!

PR #362, opened 2019-02-21
refactor websocket_wrapper.go with copyHeader func

danilvpetrov:
This PR is a follow-up of #359. 

There are similar patterns in `websocket_wrapper.go` file that can be refactored with the newly introduced `copyHeader` function.


johanbrandhorst:
THESE TESTS I BEG

johanbrandhorst:
There should be a notification for tests passing

jonny-improbable:
🙏for ✅

PR #381, opened 2019-04-06
Add grpc-web-fake-transport Package

jonny-improbable:
Migrate FakeTransport form ts-protoc-gen package and add test coverage.


integration_test/ts/_proto/improbable/grpcweb/test/test_pb.js
```diff
@@ -251,7 +251,7 @@ proto.improbable.grpcweb.test.PingRequest.prototype.getValue = function() {
 
 /** @param {string} value */
 proto.improbable.grpcweb.test.PingRequest.prototype.setValue = function(value) {
-  jspb.Message.setField(this, 1, value);
+  jspb.Message.setProto3StringField(this, 1, value);
```
easyCZ: Seems to be unrelated to the PR. Generated with a different version of proto?

jonny-improbable: confirmed generated with v1.5.0; we should, however, have better tooling around which version of protoc is used.

PR #385, opened 2019-04-11
Remove content-length header

LeonSha:
## Changes
Move the patch from envoy proxy.
https://sourcegraph.com/github.com/envoyproxy/envoy@cfc514546bc0284536893cca5fa43d7128edcd35/-/blob/source/extensions/filters/http/grpc_web/grpc_web_filter.cc#L51

## Verification
I tried the patch locally and check the content-length has been removed.

### DIFF
--- a/go/grpcweb/wrapper.go
+++ b/go/grpcweb/wrapper.go
@@ -225,6 +225,11 @@ func hackIntoNormalGrpcRequest(req *http.Request) (*http.Request, bool) {
 	}
 	req.Header.Set("content-type", strings.Replace(contentType, incomingContentType, grpcContentType, 1))
 
+	// Remove content-length header since it represents http1.1 payload size, not the sum of the h2
+	// DATA frame payload lengths. https://http2.github.io/http2-spec/#malformed This effectively
+	// switches to chunked encoding which is the default for h2
+	req.Header.Del("content-length")
+
 	return req, isTextFormat
 }
 


johanbrandhorst:
Thanks for this contribution! Is there a specific issue this fixes?

This seems to be the rationale:
```
// Remove content-length header since it represents http1.1 payload size, not the sum of the h2
// DATA frame payload lengths. https://http2.github.io/http2-spec/#malformed This effectively
// switches to chunked encoding which is the default for h2.
```

Is there an envoy issue related to this?

LeonSha:
> Thanks for this contribution! Is there a specific issue this fixes?
> 
> This seems to be the rationale:
> 
> ```
> // Remove content-length header since it represents http1.1 payload size, not the sum of the h2
> // DATA frame payload lengths. https://http2.github.io/http2-spec/#malformed This effectively
> // switches to chunked encoding which is the default for h2.
> ```
> Is there an envoy issue related to this?
https://github.com/envoyproxy/envoy/pull/2946

johanbrandhorst:
OK, so the issue is if we have a daisy chain of proxies where this proxy will transform a HTTP/1.1 request to HTTP/2 but not remove the `content-length` header, then downstream proxies that receive a HTTP/2 request with a `content-length` header might be confused.

Thanks!

johanbrandhorst:
I'll merge as soon as travis has succeeded.

PR #403, opened 2019-05-01
Don't block on close notify.

devnev:
## Changes

* Don't block on writing to close notify channel, as callers may choose to ignore this channel entirely.

## Verification

Our servers had panics from `grpc.Server`'s `GracefulStop` because there was a `serverHandlerTransport.HandleStream` blocked on the reader which was hanging as `webSocketWrappedReader.Read` was trying to write to this channel. Adding this buffer allowed the handlers to terminate.


johanbrandhorst:
Would still like a review from @marcuslongmuir

PR #411, opened 2019-05-04
Worker error "window not defined" fix

midan888:
## Changes

Added support to use in service workers

### DIFF
--- a/client/grpc-web/webpack.config.js
+++ b/client/grpc-web/webpack.config.js
@@ -23,6 +23,7 @@ module.exports = [{
       filename: `grpc-web-client.js`,
       path: DIST_DIR,
       libraryTarget: 'commonjs',
+      globalObject: 'this',
     }
   },
   {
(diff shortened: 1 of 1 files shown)


johanbrandhorst:
This is cool, what does it do? Can we test this? Can we add some documentation on how to use it?

midan888:
There is 0 configuration needed for this change, its just webpack defaults global object this as a “window”, which generates compiled code which contains window call. I use this library in service worker, which fails because of window call. For that reasons webpack has globalObject configuration to use in none window environment. When you define globalObject as “this”, it works in both environments. I tested locally, it works! I can try to think a test for this case

jonny-improbable:
This change makes sense to me (see [this article](https://mathiasbynens.be/notes/globalthis) if you want to learn more about the `this` keyword in JavaScript).

We could add a test for this, a simple integration test which consumes the generated AMD file via a service worker in any browser would do the trick; however our integration test setup is pretty gnarly. @midan888, please let me know if you want to try for this test, otherwise I'm happy for this config change to land as-is (see [webpack docs](https://webpack.js.org/configuration/output/#outputglobalobject) for confirmation).

midan888:
@jonny-improbable OK let's merge, because it will take a while to setup the test, and I really need this pr

PR #410, opened 2019-05-04
Seperate go tests from integration tests

jonny-improbable:

PR #430, opened 2019-05-10
Reinstate integration tests

jonny-improbable:
Turns out that integration tests stopped running after #381 was merged, when I split out unit and integration tests :(

Root cause is that the `integration_test` package did not have a `test` script; the default behavior of npm in this case is to exit with a status code of zero which is... annoying.

This re-instates the tests, however I have a feeling that we are going to see a lot of TLS errors as we're seeing on #410.


jonny-improbable:
Yep, sure enough, the Chrome tests are failing with the following error:

```
remote error: tls: unknown certificate authority
```

I'll try a git bisect to see if I can find which commit caused this issue to start occuring (assuming it's not a change on the BrowserStack platform itself)

jonny-improbable:
Heh, OK so the TLS errors are a red herring; the real failure was that it would not connect via the socket with the timeout window. 

Downgrading to Karma 3.1.4 resolves this issue so now I'm going to start the process of bisecting which karma version causes the failure...

jonny-improbable:
Cool, Karma 4.0.1 is the latest version we can consume; something [between the 4.0.1 and 4.1.0 releases](https://github.com/karma-runner/karma/compare/v4.0.1...v4.1.0) is breaking our tests. I'm going to merge this PR and continue investigating with the aim to raise an upstream bug report.

PR #432, opened 2019-05-11
Remove redundant integration tests

jonny-improbable:
`grpc.unary` and `grpc.invoke` both use `grpc.client` in their implementation; therefore these tests could just as easily be served through a unit test.

I'm not advocating that this PR be merged without first adding unit tests (and a unit test framework!); however I'm curious to see what the impact is on CI times.


jonny-improbable:
Closing in favor of #539 which will allow us to move many of theses tests into the `gprc-web-core` package.

PR #443, opened 2019-05-16
service worker fix for AbortController call

midan888:
window.AbortController changed to self.AbortController to work in both worker and window environments


midan888:
@jonnyreeves No, self global object exists in a browser as well https://developer.mozilla.org/en-US/docs/Web/API/Window/self.

And I don't understand why one test failed, could you help with that ?

PR #446, opened 2019-05-20
Set Content-Length in NodeHttpTransport

public:
Since this transport only supports Unary requests from the client we can easily just set the Content-Length header just before the message is written to the XHR. This avoids Node falling back to chunked encoding which is generally horrible to deal with for servers.

### DIFF
--- a/client/grpc-web-node-http-transport/src/index.ts
+++ b/client/grpc-web-node-http-transport/src/index.ts
@@ -18,6 +18,7 @@ class NodeHttp implements grpc.Transport {
   }
 
   sendMessage(msgBytes: Uint8Array) {
+    this.request.setHeader('Content-Length', msgBytes.byteLength);
     this.request.write(toBuffer(msgBytes));
     this.request.end();
   }


johanbrandhorst:
Dupe of #427 unfortunately, but we probably should just merge that.

johanbrandhorst:
Reopening this because I realised #427 has not seen any replies from the contributor for a while. The same comments apply to this change though:

> If I understand this change correctly then the content-length header should only set for unary requests. If that's correct please can you make this change? FYI, you can determine the request type from via the [options member](https://github.com/improbable-eng/grpc-web/blob/master/client/grpc-web-node-http-transport/src/index.ts#L13) property.

> It would also be great if you could look to add a unit test for this option; however I appreciate that this package has not test framework - however I would be happy to help guide you on how to contribute such a test if you have the time and interest.

johanbrandhorst:
@public If you're interested in addressing these concerns, I can get this merged.

public:
I actually asked Michael to make that PR :) I'll catch up with him.

jonny-improbable:
@public I'm going to close this PR as it is essentially a dupe of #427 and from your comments it appears that you are in contact with the author of that PR.

Issue #453, opened 2019-05-24
Allow base path before service-name and method


mangas:
**Versions of relevant software used**
Latest
**What happened**
Addding the wrapper to an existing router under a base path requires fixing the path before passing the request to the wrapper otherwise service is not found
**What you expected to happen**
That the last 2 parts of the path were used to match registered services rather than the full path
**How to reproduce it (as minimally and precisely as possible)**:
```
router.New()
r.POST("/grpc/*all", wrappedServer)

r.ServeHttp(resp,req)
```

johanbrandhorst:
I don't think it's unreasonable for us to expect the user to use [http.StripPrefix](https://golang.org/pkg/net/http/#StripPrefix) when serving on non-root paths. We could probably make this more clear in the documentation though. Would you be interested in contributing a documentation fix for this?

mangas:
What is the downside of handling gracefully on the wrapper side?

johanbrandhorst:
I don't think it's as simple as just taking the last two elements, always. I would prefer not to add special case handling in the wrapper when it only affects the small portion of users who do not serve on `/`.

mangas:
What are the cases when it's not just taking the last 2 elements?

johanbrandhorst:
According to that exact doc, it specifies the `:path` prefix to the `Service-Name` / `MethodName` tuple, AND it also states that sometimes the path is not in this form, but that it is discouraged. It doesn't sound like something that we want to rely in, again, _especially_ when this is so easily handed in the initialization.

Even if that were not the case, consider that this would introduce a mandatory parsing step of the URL path into the wrapper, which has a cost to all users.

I think we need much more convincing evidence of this being widespread and painful to workaround before we should add it on the wrapper.

mangas:
Following the current pattern for cors and origin, this could be introduced as a func for the parsing, allowing to switch between the behaviour hence not affecting everyone but still providing it as part of the wrapper?

johanbrandhorst:
That would have less of an impact, but it still adds to the API surface and maintenance burden, and I don't understand why `http.StripPrefix` isn't suitable for your use case? Maybe if you could share a bit more about your setup we could come to a compromise.

mangas:
httpStripPrefix does work for my case but is something that basically needs to be kind of discovered and implemented for each project and that doesn't actually follow the protocol as would be expected. So it's more of a question of where this should be handled. Since the library does use the path to match registered services it seems to me like it would be the natural place. 

[...]

PR #454, opened 2019-05-24
Allow non root resources

mangas:
https://github.com/improbable-eng/grpc-web/issues/453

## Changes
Added option `allowNonRootResources`
Added a function `WrappedGrpcServer` that encapsulates the logic of getting the gRPC endpoint from http.Request

## Verification
Added Unit-Tests that ensure the same behavior for the canonical case 
Tested with flag on by default

### DIFF
--- a/go/grpcweb/helpers.go
+++ b/go/grpcweb/helpers.go
@@ -40,7 +40,7 @@ func WebsocketRequestOrigin(req *http.Request) (string, error) {
 	return parsed.Host, nil
 }
 
-func GetGRPCEndpoint(req *http.Request) string {
+func getGRPCEndpoint(req *http.Request) string {
 	endpoint := pathMatcher.FindString(strings.TrimRight(req.URL.Path, "/"))
 	if len(endpoint) == 0 {
 		return req.URL.Path
--- a/go/grpcweb/helpers.go
+++ b/go/grpcweb/helpers.go
@@ -7,9 +7,13 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"regexp"
+	"strings"
 
 	"google.golang.org/grpc"
 )
+
+var pathMatcher = regexp.MustCompile(`/[^/]*/[^/]*$`)
 
 // ListGRPCResources is a helper function that lists all URLs that are registered on gRPC server.
 //
@@ -34,4 +38,13 @@ func WebsocketRequestOrigin(req *http.Request) (string, error) {
 		return "", fmt.Errorf("failed to parse url for grpc-websocket origin check: %q. error: %v", origin, err)
 	}
 	return parsed.Host, nil
+}
+
+func GetGRPCEndpoint(req *http.Request) string {
+	endpoint := pathMatcher.FindString(strings.TrimRight(req.URL.Path, "/"))
+	if len(endpoint) == 0 {
+		return req.URL.Path
+	}
+
+	return endpoint
 }
--- a/go/grpcweb/helpers_test.go
+++ b/go/grpcweb/helpers_test.go
@@ -4,7 +4,6 @@
 package grpcweb_test
 
 import (
-	"net/http/httptest"
 	"sort"
 	"testing"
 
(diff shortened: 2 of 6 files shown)


go/grpcweb/helpers.go
```diff
@@ -35,3 +39,12 @@ func WebsocketRequestOrigin(req *http.Request) (string, error) {
 	}
 	return parsed.Host, nil
 }
+
+func GetGRPCEndpoint(req *http.Request) string {
```
johanbrandhorst: Do we need to export this function?

go/grpcweb/helpers_test.go
```diff
@@ -34,3 +35,26 @@ func TestListGRPCResources(t *testing.T) {
...
+	}
+
+	for _, c := range cases {
+		req := httptest.NewRequest("GET", c.input, nil)
+
+		result := grpcweb.GetGRPCEndpoint(req)
+
+		assert.Equal(t, c.output, result)
+	}
+
```
johanbrandhorst: Nitpick: remove this newline.

go/grpcweb/wrapper.go
```diff
@@ -56,13 +57,20 @@ func WrapServer(server *grpc.Server, options ...Option) *WrappedGrpcServer {
 	if websocketOriginFunc == nil {
 		websocketOriginFunc = defaultWebsocketOriginFunc
 	}
+
+	endpointFunc := func(req *http.Request) string { return req.URL.Path }
```
johanbrandhorst: Can you write this over 3 lines please.

go/grpcweb/helpers.go
mangas: We don't :)

mangas: Actually the reason I exported it is because the test runs from outside the package, I could either make it internal or create an internal test file?

mangas:
Is it worth running one of the suites with the non default option to make sure that there is always something that tests the option?

go/grpcweb/helpers.go
johanbrandhorst: Create an internal test file please

mangas: Do you have a preference for this one? 

1) move the test to a helpers_internal_test
2) change helpers_test.go package to grpcweb (remove _test)
3) leave exported

johanbrandhorst:
> Is it worth running one of the suites with the non default option to make sure that there is always something that tests the option?

I'd rather add a new test which exercises the option, but if it's complicated it's not necessary. it's a pretty straightforward change.

mangas:
@johanbrandhorst @jonnyreeves I believe the failed tests are unrelated but let me know if there is anything I'm missing here

johanbrandhorst:
Restarted it

mangas:
Will we be able to get some more 👀 on this PR? 🙏

go/grpcweb/options.go
```diff
@@ -99,3 +101,9 @@ func WithWebsocketOriginFunc(websocketOriginFunc func(req *http.Request) bool) O
 		o.websocketOriginFunc = websocketOriginFunc
 	}
 }
+
+func WithAllowNonRootResource(allowNonRootResources bool) Option {
```
johanbrandhorst: Ah, something I missed before, please add a comment explaining what this is useful, ideally with an example.

mangas: Will get on that shortly 👍

mangas: ^^

go/grpcweb/options.go
```diff
@@ -99,3 +101,15 @@ func WithWebsocketOriginFunc(websocketOriginFunc func(req *http.Request) bool) O
 		o.websocketOriginFunc = websocketOriginFunc
 	}
 }
+
+// WithAllowNonRootResource enables the gRPC wrapper to serve requests that have a path prefix
+// added to the URL, before the service name and method placeholders.
+//
+// This should be set to false when exposing the endpoint as the root resource, to avoid path processing.
+//
+// The default behaviour is `false`, i.e. always serves requests assuming there is no prefix to the gRPC endpoint.
```
johanbrandhorst: Maybe just add that enabling this incurs a slight performance cost since we run a regex on each request.

mangas: // This should be set to false when exposing the endpoint as the root resource, to avoid path processing.

johanbrandhorst: Yeah just make it a little more explicit that there is a performance cost.

johanbrandhorst: I mean, rather than saying "leave this off", say "enabling this is not for free".

johanbrandhorst: (not literally :P)

johanbrandhorst:
:crossed_fingers: lets hope tests pass

johanbrandhorst:
:repeat:

mangas:
🔁

johanbrandhorst:
:roll_eyes: :recycle:

johanbrandhorst:
Thanks for your contribution!

mangas:
Thank you for the help 👍

PR #468, opened 2019-06-07
build: Use prototool for installing protoc and compiling proto files

jonny-improbable:
## Changes
- Adds prototool configuration
- Update contributing guide
- Update travis script
- Removes yolo bash scripts
- Upgrades `google-protobuf` dependency to keep the protoc version and npm dependency in sync (3.8.0)

## Verification
- This build will pass 🤞


install-prototool.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+# Installs prototool; used for managing protoc the installation and compilation flags.
+# This script is intended to be run by Travis CI
+
+set -ex
+
+curl -sSL \
+  https://github.com/uber/prototool/releases/download/v1.7.0/prototool-$(uname -s)-$(uname -m) \
```
johanbrandhorst: Do we want to parametrise the version as well?

integration_test/proto/prototool.yaml
```diff
@@ -0,0 +1,23 @@
...
+generate:
+  go_options:
+    # The base import path. This should be the go path of the prototool.yaml file.
+    # This is required if you have any go plugins.
+    import_path: integration_test/proto
+
+  plugins:
+    - name: go
+      type: go
+      output: ../go/_proto
```
johanbrandhorst: I think you want to make this generate the gRPC server too:
```yaml
flags: plugin=grpc
```

johanbrandhorst:
Looks like we're missing the generated typescript service too.

install-prototool.sh
jonny-improbable: Aye, probably should host it into the travis.yaml - also my build is failing, presumably because I can't write to `/usr/local/bin` on a travis host.

jonny-improbable:
@johanbrandhorst looks like this is now ready for review :)

jonny-improbable:
> What happened to integration_test/ts/_proto/improbable/grpcweb/test/test_pb_service.ts?

This was an artefact from a (very!) old version to protoc-gen-ts where is used to output a single `.ts` source file for services. Newer versions generate 2 source files: a javascript file and a typescript definition file (`.d.ts`).

TL;DR - this file was redundant and not used.

PR #470, opened 2019-06-07
Fail on lint issues and fixup the ones that have crept in

### LINKED ISSUE #469: Linting issues are not failing the CI Builds
Lint failures [are occuring during CI Builds on master](https://travis-ci.org/improbable-eng/grpc-web/jobs/542169483#L5890); however travis is not reporting build failure :(

jonny-improbable:
fixes #469

### DIFF
--- a/lint-all.sh
+++ b/lint-all.sh
@@ -1,8 +1,9 @@
 #!/bin/bash
+
 set -e
 
 echo "Linting go sources"
-cd go && ./lint.sh && cd ..
+cd go && . ./lint.sh && cd ..
 
 echo "Linting TypeScript sources"
 npm run lint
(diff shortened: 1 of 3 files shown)


lint-all.sh
```diff
@@ -1,8 +1,9 @@
 #!/bin/bash
+
 set -e
 
 echo "Linting go sources"
-cd go && ./lint.sh && cd ..
+cd go && . ./lint.sh && cd ..
```
johanbrandhorst: This confuses me. Do we have to source the script to run it? Maybe we should rewrite the script?

jonny-improbable:
If I don't source it, a non-zero exit code will be ignored. TBH my bash-fu
is pretty weak, but this worked so I was happy (:

PR #406, opened 2019-05-03
Add ARM binaries to release script

jonny-improbable:
Fixes #391

PR #521, opened 2019-08-05
Add ARM build to release script

johanbrandhorst:
Should make the release process even easier.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,7 +13,7 @@ export GOPATH=$(pwd)
 
 # Checkout project sources into your new go workspace
 go get -u github.com/improbable-eng/grpc-web/go
-cd $GOPATH/src/github.com/improbable-eng/grpc-web 
+cd $GOPATH/src/github.com/improbable-eng/grpc-web
 
 # Install go dependencies
 dep ensure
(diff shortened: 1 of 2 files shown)


jonny-improbable:
See https://github.com/improbable-eng/grpc-web/pull/406; as mentioned in that thread, you can't cross compile for ARM on a macOS host, which, unfortunately, is what I use to build and publish releases.

We should either
1. Build release binaries a docker container
2. Automate the entire release process through a CI job

Option 1 is certainly simpler (:

CONTRIBUTING.md
```diff
@@ -74,5 +74,6 @@ This is your prompt to open a web browser on https://localhost:9876 at which poi
 1. From a fresh checkout of master, create a release branch, ie: `feature/prepare-x.y.z-release`
 2. Update `CHANGELOG.md` by comparing commits to master since the last Github Release
 3. Raise a pull request for your changes, have it reviewed and landed into master.
-4. Switch your local checkout back to the master branch, pull your merged changes and run `./publish-release.sh`.
-5. Create the ARM binaries and attach them to the Github release.
\ No newline at end of file
+4. Make the release on github using the new commit in master.
```
jonny-improbable: The `publish-release.sh` script takes care of creating the github release and attaching the generated binaries to it. It also publishes the client(s) to npm.

johanbrandhorst: whew I clearly misunderstood the depth of this script.

johanbrandhorst:
Ah, doh! I see. Nevermind this. I set up a auto publishing job for the grpc-gateway but we were using CircleCI. Might be possible to copy some of that though: https://github.com/grpc-ecosystem/grpc-gateway/blob/master/.circleci/config.yml#L77

PR #527, opened 2019-08-06
Revert "Check flusher interface before calling Flush (#479)"

johanbrandhorst:
This reverts commit cf7d15a039114df2a0e3a18002705bb67fdd5167.
This introduced a recursive call to `Flush` in the proxy, which
will crash the application.

### DIFF
--- a/go/grpcweb/grpc_web_response.go
+++ b/go/grpcweb/grpc_web_response.go
@@ -59,15 +59,10 @@ func (w *grpcWebResponse) WriteHeader(code int) {
 }
 
 func (w *grpcWebResponse) Flush() {
-	f, ok := w.wrapped.(http.Flusher)
-	if !ok {
-		return
-	}
-
 	if w.wroteHeaders || w.wroteBody {
 		// Work around the fact that WriteHeader and a call to Flush would have caused a 200 response.
 		// This is the case when there is no payload.
-		f.Flush()
+		w.wrapped.(http.Flusher).Flush()
 	}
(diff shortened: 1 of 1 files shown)


go/grpcweb/grpc_web_response.go
```diff
@@ -95,7 +90,7 @@ func (w *grpcWebResponse) finishRequest(req *http.Request) {
 		w.copyTrailersToPayload()
 	} else {
 		w.WriteHeader(http.StatusOK)
-		w.Flush()
```
jonny-improbable: @johanbrandhorst whilst we wait for CI to go green on this PR...

...isn't this the root cause of the issue? Shouldn't this be replaced with a safe up-cast on `w.wrapped` before calling `Flush()`?

jonny-improbable:
Given this failure has had an impact on consumers, it implies we are missing a test. @light0x, would you be in a position to help us contribute a test-case which would help us catch this regression in future, please?

johanbrandhorst:
Rerunning failed tests

go/grpcweb/grpc_web_response.go
johanbrandhorst: I haven't looked at it in detail, lets discuss after merging the fix.

jonny-improbable:
Given that the 2 failures occurred on older versions of Chrome I am happy to chalk the failure down to BrowerStack flakiness and will use Admin powers to merge.

PR #539, opened 2019-08-31
Split out core package

jonny-improbable:
As discussed, split up the `client/gprc-web` package into smaller discrete chunks in order to allow easier extension without polluting the 'consumer' `grpc` namespace.

New packages:
* *gprc-web-core* -- All of `grpc-web` without any transports
* *gprc-web-websocket-transport* -- The websocket transport we used to ship in `grpc-web`, consumers will need to depend on this transport after this change (breaking!)
* *grpc-web-cross-browser-transport* -- Fetch/XHR based transports and the elusive `DefaultXhrTransport` class that others want to extend.

This leaves the `grpc-web` package pretty empty, just an index file which exports the `grpc` namespace and configured the `defaultTransportFactory` to be the cross-browser one.


client/grpc-web-core/src/transport.ts
```diff
@@ -0,0 +1,51 @@
...
+    sendMessage(_: Uint8Array): void {
+        throw new Error("No transport configured")
+    }
+
+    start(_: Metadata): void {
+        throw new Error("No transport configured")
+    }
+}
+
+let defaultTransportFactory: TransportFactory = (_: TransportOptions) => new NullTransport();
```
jonny-improbable: as discussed, this should become a "true" global at some point, but let's do that in another PR.

client/grpc-web-cross-browser-transport/tsconfig.json
```diff
@@ -0,0 +1,27 @@
+{
```
MarcusLongmuir: You can use `extends` (https://www.typescriptlang.org/docs/handbook/tsconfig-json.html) to use a common base `tsconfig.json`, but this might be against the principle of splitting these up.

client/grpc-web-core/src/transport.ts
jonny-improbable: Actually, we should make this change in this PR to avoid issues with consuming a webpack'd build; somehting like https://github.com/ungap/global-this seems appropriate.

Issue #538, opened 2019-08-30
(Docs) Broken link to example repo


jbpratt78:
**What happened**
404 not found
**What you expected to happen**
A beautiful example would appear
**How to reproduce it (as minimally and precisely as possible)**:
[Click here](https://github.com/improbable-eng/grpc-web/tree/master/example)
**Full logs to relevant components**
https://github.com/improbable-eng/grpc-web/blob/master/client/grpc-web/README.md has an example link that is broken

johanbrandhorst:
Hi @jbpratt78, thanks for the bug report! Would you be interested in contributing a fix for this? It should be an easy fix :).

jbpratt78:
Sure, if it should just be swapped for the React example link, then I will do that now! :+1:

PR #540, opened 2019-08-31
updated example link in docs

jbpratt78:
## Changes
Swapped old example link to point to the React example project as per #538 

## Verification
https://github.com/jbpratt78/grpc-web/tree/docs-example-link-fix/client/grpc-web :)


johanbrandhorst:
Thanks!

PR #546, opened 2019-09-12
Increase websocket timeout to send ping to adapt nginx's websocket proxy

angwangiot:
当我尝试用nginx代理grpc-web中的websocket时，由于nginx有检测断开机制，导致超过60秒没有推送消息会断开。
nginx官网的解释时这样的：http://nginx.org/en/docs/http/websocket.html

> When I try to use nginx to proxy the websocket in grpc-web, since nginx has a detection disconnection mechanism, no push message will be disconnected for more than 60 seconds.
Nginx official website explains this

## Changes

根据nginx的官网说明，我在服务器的推送上增加了超时发送Ping的功能。

> According to the official website of nginx, I added the function of timeout to send Ping on the push of the server.

## Verification

经过修改，已经可以实现nginx代理websocket并且不自动断开了。

> After modification, the nginx proxy websocket can be implemented and not automatically disconnected.

Sorry, My English is not good, so I did not use English instructions.

(Translations courtesy of google translate - @johanbrandhorst)

### DIFF
--- a/go/grpcweb/options.go
+++ b/go/grpcweb/options.go
@@ -19,6 +19,7 @@ type options struct {
 	corsForRegisteredEndpointsOnly bool
 	originFunc                     func(origin string) bool
 	enableWebsockets               bool
+	websocketPingTimeout           uint
 	websocketOriginFunc            func(req *http.Request) bool
 	allowNonRootResources          bool
 }
--- a/go/grpcweb/options.go
+++ b/go/grpcweb/options.go
@@ -3,7 +3,10 @@
 
 package grpcweb
 
-import "net/http"
+import (
+	"net/http"
+	"time"
+)
 
 var (
 	defaultOptions = &options{
@@ -19,7 +22,7 @@ type options struct {
 	corsForRegisteredEndpointsOnly bool
 	originFunc                     func(origin string) bool
 	enableWebsockets               bool
-	websocketPingTimeout           uint
+	websocketPingInterval          time.Duration
 	websocketOriginFunc            func(req *http.Request) bool
 	allowNonRootResources          bool
 }
@@ -93,12 +96,12 @@ func WithWebsockets(enableWebsockets bool) Option {
 	}
 }
 
-// WithWebsocketPingTimeout allows for websocket send ping.
+// WithWebsocketPingInterval enables websocket keepalive pinging with the configured timeout (in seconds).
 //
-// The default behaviour is 0, i.e. to disallow websocket ping
-func WithWebsocketPingTimeout(websocketPingTimeout uint) Option {
+// The default behaviour is to disable websocket pinging.
+func WithWebsocketPingInterval(websocketPingInterval time.Duration) Option {
 	return func(o *options) {
-		o.websocketPingTimeout = websocketPingTimeout
+		o.websocketPingInterval = websocketPingInterval
 	}
 }
 
--- a/go/grpcweb/options.go
+++ b/go/grpcweb/options.go
@@ -96,7 +96,7 @@ func WithWebsockets(enableWebsockets bool) Option {
 	}
 }
 
-// WithWebsocketPingInterval enables websocket keepalive pinging with the configured timeout (in seconds).
+// WithWebsocketPingInterval enables websocket keepalive pinging with the configured timeout.
 //
 // The default behaviour is to disable websocket pinging.
 func WithWebsocketPingInterval(websocketPingInterval time.Duration) Option {
--- a/go/grpcweb/options.go
+++ b/go/grpcweb/options.go
@@ -93,6 +93,15 @@ func WithWebsockets(enableWebsockets bool) Option {
 	}
 }
 
+// WithWebsocketPingTimeout allows for websocket send ping.
+//
+// The default behaviour is 0, i.e. to disallow websocket ping
+func WithWebsocketPingTimeout(websocketPingTimeout uint) Option {
+	return func(o *options) {
+		o.websocketPingTimeout = websocketPingTimeout
+	}
+}
+
 // WithWebsocketOriginFunc allows for customizing the acceptance of Websocket requests - usually to check that the origin
 // is valid.
 //
--- a/go/grpcweb/websocket_wrapper.go
+++ b/go/grpcweb/websocket_wrapper.go
@@ -45,6 +45,7 @@ func (w *webSocketResponseWriter) ping() {
 		close(dispose)
 		return nil
 	})
+	defer ticker.Stop()
 	for {
 		select {
 		case <-dispose:
--- a/go/grpcweb/websocket_wrapper.go
+++ b/go/grpcweb/websocket_wrapper.go
@@ -10,21 +10,20 @@ import (
 	"net/http"
 	"net/textproto"
 	"strings"
-	"sync"
 	"time"
 
+	"github.com/desertbit/timer"
 	"github.com/gorilla/websocket"
 	"golang.org/x/net/http2"
 )
 
 type webSocketResponseWriter struct {
-	writtenHeaders      bool
-	wsConn              *websocket.Conn
-	headers             http.Header
-	flushedHeaders      http.Header
-	timeOutInterval     time.Duration
-	lastMessageTime     time.Time
-	lastMessageTimeLock *sync.Mutex
+	writtenHeaders  bool
+	wsConn          *websocket.Conn
+	headers         http.Header
+	flushedHeaders  http.Header
+	timeOutInterval time.Duration
+	timer           *timer.Timer
 }
 
 func newWebSocketResponseWriter(wsConn *websocket.Conn) *webSocketResponseWriter {
@@ -38,7 +37,7 @@ func newWebSocketResponseWriter(wsConn *websocket.Conn) *webSocketResponseWriter
 
 func (w *webSocketResponseWriter) enablePing(timeOutInterval time.Duration) {
 	w.timeOutInterval = timeOutInterval
-	w.lastMessageTimeLock = &sync.Mutex{}
+	w.timer = timer.NewTimer(w.timeOutInterval)
 	dispose := make(chan bool)
 	w.wsConn.SetCloseHandler(func(code int, text string) error {
 		close(dispose)
@@ -51,23 +50,14 @@ func (w *webSocketResponseWriter) ping(dispose chan bool) {
 	if dispose == nil {
 		return
 	}
-	ticker := time.NewTicker(w.timeOutInterval)
-	defer ticker.Stop()
+	defer w.timer.Stop()
 	for {
 		select {
 		case <-dispose:
 			return
-		case t := <-ticker.C:
-			var sendPing = false
-			w.lastMessageTimeLock.Lock()
-			if t.After(w.lastMessageTime.Add(w.timeOutInterval)) {
-				sendPing = true
-				w.lastMessageTime = time.Now()
-			}
-			w.lastMessageTimeLock.Unlock()
-			if sendPing {
-				w.wsConn.WriteMessage(websocket.PingMessage, []byte{})
-			}
+		case <-w.timer.C:
+			w.timer.Reset(w.timeOutInterval)
+			w.wsConn.WriteMessage(websocket.PingMessage, []byte{})
 		}
 	}
 }
@@ -80,10 +70,8 @@ func (w *webSocketResponseWriter) Write(b []byte) (int, error) {
 	if !w.writtenHeaders {
 		w.WriteHeader(http.StatusOK)
 	}
-	if w.timeOutInterval > time.Second {
-		w.lastMessageTimeLock.Lock()
-		w.lastMessageTime = time.Now()
-		w.lastMessageTimeLock.Unlock()
+	if w.timeOutInterval > time.Second && w.timer != nil {
+		w.timer.Reset(w.timeOutInterval)
 	}
 	return len(b), w.wsConn.WriteMessage(websocket.BinaryMessage, b)
 }
--- a/go/grpcweb/websocket_wrapper.go
+++ b/go/grpcweb/websocket_wrapper.go
@@ -58,7 +58,6 @@ func (w *webSocketResponseWriter) ping() {
 			}
 		}
 	}
-Stop:
 	ticker.Stop()
 }
 
--- a/go/grpcweb/websocket_wrapper.go
+++ b/go/grpcweb/websocket_wrapper.go
@@ -37,7 +37,7 @@ func newWebSocketResponseWriter(wsConn *websocket.Conn) *webSocketResponseWriter
 }
 
 func (w *webSocketResponseWriter) EnablePing(timeOutInterval uint) {
-	if timeOutInterval > 1 {
+	if timeOutInterval >= 1 {
 		w.timeOutInterval = timeOutInterval
 	} else {
 		w.timeOutInterval = 30
@@ -61,11 +61,13 @@ func (w *webSocketResponseWriter) ping() {
 			return
 		case <-ticker.C:
 			w.tickerCountLock.Lock()
-			defer w.tickerCountLock.Unlock()
 			w.tickerCount++
 			if w.tickerCount >= w.timeOutInterval {
 				w.tickerCount = 0
+				w.tickerCountLock.Unlock()
 				w.wsConn.WriteMessage(websocket.PingMessage, []byte{})
+			} else {
+				w.tickerCountLock.Unlock()
 			}
 		}
 	}
@@ -80,8 +82,8 @@ func (w *webSocketResponseWriter) Write(b []byte) (int, error) {
 		w.WriteHeader(http.StatusOK)
 	}
 	w.tickerCountLock.Lock()
-	defer w.tickerCountLock.Unlock()
 	w.tickerCount = 0
+	w.tickerCountLock.Unlock()
 	return len(b), w.wsConn.WriteMessage(websocket.BinaryMessage, b)
 }
 
--- a/go/grpcweb/websocket_wrapper.go
+++ b/go/grpcweb/websocket_wrapper.go
@@ -1,7 +1,6 @@
 package grpcweb
 
 import (
-	"sync"
 	"bufio"
 	"bytes"
 	"context"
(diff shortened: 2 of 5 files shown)


go/grpcweb/websocket_wrapper.go
```diff
@@ -10,25 +10,55 @@ import (
...
+	dispose := make(chan bool)
+	ticker := time.NewTicker(time.Second)
+	w.wsConn.SetCloseHandler(func(code int, text string) error {
+		close(dispose)
+		return nil
+	})
+	for {
+		select {
+		case <-dispose:
+			goto Stop
```
johanbrandhorst: Lets just return here.
```suggestion
			return
```

go/grpcweb/websocket_wrapper.go
```diff
@@ -10,25 +10,55 @@ import (
...
+}
+
+func (w *webSocketResponseWriter) ping() {
+	dispose := make(chan bool)
+	ticker := time.NewTicker(time.Second)
+	w.wsConn.SetCloseHandler(func(code int, text string) error {
+		close(dispose)
+		return nil
+	})
+	for {
```
johanbrandhorst: Defer `stop`
```suggestion
	defer ticker.Stop()
	for {
```

go/grpcweb/websocket_wrapper.go
```diff
@@ -10,25 +10,55 @@ import (
...
+			goto Stop
+		case <-ticker.C:
+			w.tickerCount++
+			if w.tickerCount >= w.timeOutInterval {
+				w.tickerCount = 0
+				w.wsConn.WriteMessage(websocket.PingMessage, []byte{})
+			}
+		}
 	}
+Stop:
```
johanbrandhorst: Remove this
```suggestion
```

go/grpcweb/websocket_wrapper.go
```diff
@@ -10,25 +10,55 @@ import (
...
+	return wsRW
+}
+
+func (w *webSocketResponseWriter) ping() {
+	dispose := make(chan bool)
+	ticker := time.NewTicker(time.Second)
+	w.wsConn.SetCloseHandler(func(code int, text string) error {
+		close(dispose)
+		return nil
+	})
```
johanbrandhorst: This has to be set in the `newWebSocketResponseWriter` or it will be racy.

go/grpcweb/websocket_wrapper.go
```diff
@@ -39,6 +69,7 @@ func (w *webSocketResponseWriter) Write(b []byte) (int, error) {
 	if !w.writtenHeaders {
 		w.WriteHeader(http.StatusOK)
 	}
+	w.tickerCount = 0
```
johanbrandhorst: We need a mutex around this variable

go/grpcweb/websocket_wrapper.go
```diff
@@ -10,24 +11,67 @@ import (
...
-	writtenHeaders bool
-	wsConn         *websocket.Conn
-	headers        http.Header
-	flushedHeaders http.Header
+	writtenHeaders  bool
+	wsConn          *websocket.Conn
+	headers         http.Header
+	flushedHeaders  http.Header
+	timeOutInterval uint
+	tickerCount     uint
```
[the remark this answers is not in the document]
angwangiot: 您好，我尝试使用lastMessageTime代替tickerCount来处理校验是否超时发送Ping。但是使用ticker的思路导致会频繁调用ticker的超时函数进行额外的计算，而且精度会受ticker间隔大小的影响，所以我尝试了使用timer.Reset()方法来重置定时器。目前看起来工作是正常的，但是不知道这样实现是否有未考虑到的隐患。
Hello, I am trying to use lastMessageTime instead of tickerCount to handle whether the check is timed to send a ping. But the idea of using ticker leads to frequent calls to ticker's timeout function for additional calculations, and the precision is affected by the size of the ticker interval, so I tried to reset the timer using the timer.Reset() method. At the moment it seems that the work is normal, but I don't know if there are hidden dangers in this implementation.
（Translations courtesy of google translate）

go/grpcweb/websocket_wrapper.go
```diff
@@ -10,16 +10,19 @@ import (
...
-	writtenHeaders bool
-	wsConn         *websocket.Conn
-	headers        http.Header
-	flushedHeaders http.Header
+	writtenHeaders  bool
+	wsConn          *websocket.Conn
+	headers         http.Header
+	flushedHeaders  http.Header
+	timeOutInterval time.Duration
+	timer           *time.Timer
```
johanbrandhorst: I think we want a  `ticker`, not a `timer`. `time.Ticker`.
```suggestion
	ticker           *time.Ticker
```

johanbrandhorst: We also need a `lastMessageTime time.Time` so that in the channel loop we can compare `if tick.After(w.lastMessageTime.Add(w.timeOutInterval))`.

go/grpcweb/websocket_wrapper.go
```diff
@@ -31,6 +34,36 @@ func newWebSocketResponseWriter(wsConn *websocket.Conn) *webSocketResponseWriter
...
+func (w *webSocketResponseWriter) ping(dispose chan bool) {
+	if dispose == nil {
+		return
+	}
+	defer w.timer.Stop()
+	for {
+		select {
+		case <-dispose:
+			return
+		case <-w.timer.C:
```
johanbrandhorst: Capture the time here:
```suggestion
		case tickTime := <-w.ticker.C:
```

go/grpcweb/websocket_wrapper.go
```diff
@@ -31,6 +34,36 @@ func newWebSocketResponseWriter(wsConn *websocket.Conn) *webSocketResponseWriter
...
+		return nil
+	})
+	go w.ping(dispose)
+}
+
+func (w *webSocketResponseWriter) ping(dispose chan bool) {
+	if dispose == nil {
+		return
+	}
+	defer w.timer.Stop()
```
johanbrandhorst:
```suggestion
	defer w.ticker.Stop()
```

go/grpcweb/websocket_wrapper.go
```diff
@@ -31,6 +34,36 @@ func newWebSocketResponseWriter(wsConn *websocket.Conn) *webSocketResponseWriter
 	}
 }
 
+func (w *webSocketResponseWriter) EnablePing(timeOutInterval time.Duration) {
+	if timeOutInterval < time.Second {
+		return
+	}
+	w.timeOutInterval = timeOutInterval
+	w.timer = time.NewTimer(w.timeOutInterval)
```
johanbrandhorst:
```suggestion
	w.ticker = time.NewTicker(w.timeOutInterval)
```

go/grpcweb/websocket_wrapper.go
```diff
@@ -31,6 +34,36 @@ func newWebSocketResponseWriter(wsConn *websocket.Conn) *webSocketResponseWriter
 	}
 }
 
+func (w *webSocketResponseWriter) EnablePing(timeOutInterval time.Duration) {
+	if timeOutInterval < time.Second {
+		return
+	}
```
johanbrandhorst: We don't need this since we're checking the duration in the input.

go/grpcweb/websocket_wrapper.go
```diff
@@ -31,6 +34,36 @@ func newWebSocketResponseWriter(wsConn *websocket.Conn) *webSocketResponseWriter
 	}
 }
 
+func (w *webSocketResponseWriter) EnablePing(timeOutInterval time.Duration) {
```
johanbrandhorst: Lets unexport this function, I don't think it needs to be exported.
```suggestion
func (w *webSocketResponseWriter) enablePing(timeOutInterval time.Duration) {
```

go/grpcweb/websocket_wrapper.go
```diff
@@ -31,6 +34,36 @@ func newWebSocketResponseWriter(wsConn *websocket.Conn) *webSocketResponseWriter
...
+	if dispose == nil {
+		return
+	}
+	defer w.timer.Stop()
+	for {
+		select {
+		case <-dispose:
+			return
+		case <-w.timer.C:
+			w.timer.Reset(w.timeOutInterval)
```
johanbrandhorst: Here we can check if the `lastMessageTime` is more than `timeOutInterval` since.
```suggestion
			if tickTime.After(w.lastMessageTime.Add(w.timeOutInterval)) {
```

go/grpcweb/websocket_wrapper.go
```diff
@@ -31,6 +34,36 @@ func newWebSocketResponseWriter(wsConn *websocket.Conn) *webSocketResponseWriter
...
+			w.wsConn.WriteMessage(websocket.PingMessage, []byte{})
```
johanbrandhorst:
```suggestion
				w.wsConn.WriteMessage(websocket.PingMessage, []byte{})
			}
```

go/grpcweb/websocket_wrapper.go
johanbrandhorst: And we need a mutex for `lastMessageTime`.

go/grpcweb/websocket_wrapper.go
```diff
@@ -39,6 +72,9 @@ func (w *webSocketResponseWriter) Write(b []byte) (int, error) {
 	if !w.writtenHeaders {
 		w.WriteHeader(http.StatusOK)
 	}
+	if w.timeOutInterval > time.Second && w.timer != nil {
+		w.timer.Reset(w.timeOutInterval)
+	}
```
johanbrandhorst: This just becomes a matter of setting `lastMessageTime`
```suggestion
	w.tmu.Lock()
	w.lastMessageTime = time.Now()
	w.tmu.Unlock()
```

go/grpcwebproxy/main.go
```diff
@@ -38,7 +38,8 @@ var (
 	runHttpServer = pflag.Bool("run_http_server", true, "whether to run HTTP server")
 	runTlsServer  = pflag.Bool("run_tls_server", true, "whether to run TLS server")
 
-	useWebsockets = pflag.Bool("use_websockets", false, "whether to use beta websocket transport layer")
+	useWebsockets         = pflag.Bool("use_websockets", false, "whether to use beta websocket transport layer")
+	websocketPingInterval = pflag.Duration("websocket_ping_interval", 0, "whether to use websocket keepalive pinging. Only used when using websockets.")
```
johanbrandhorst:
```suggestion
	websocketPingInterval = pflag.Duration("websocket_ping_interval", 0, "whether to use websocket keepalive pinging. Only used when using websockets. Configured interval must be >= 1s")
```

go/grpcwebproxy/main.go
```diff
@@ -71,6 +72,13 @@ func main() {
 			grpcweb.WithWebsockets(true),
 			grpcweb.WithWebsocketOriginFunc(makeWebsocketOriginFunc(allowedOrigins)),
 		)
+		if *websocketPingInterval >= time.Second {
+			logrus.Println(fmt.Sprintf("enables websocket keepalive pinging, the timeout interval is %f seconds", websocketPingInterval.Seconds()))
```
johanbrandhorst:
```suggestion
			logrus.Println(fmt.Sprintf("websocket keepalive pinging enabled, the timeout interval is %s", websocketPingInterval.String()))
```

go/grpcweb/websocket_wrapper.go
angwangiot: 我想，我应该理解了您之前的建议，也确实这么做了（可以参考下面的websocket_wrapper.go未提交之前的修改内容）。
只是在改完之后有了其他的想法，导致我又进行了其他的改动。改动的原因是这样的：
我的意思是，用ticker那样做确实没问题，但是这样有两个地方不是很满意。第一，ticker会在每秒都触发然后进行一次计算，这样会进行比较多的无用的计算。比如，我设置间隔为30秒，那么，ticker进行了30次计算，只有最后一次的计算才有效，除了最后一次的计算，其他的计算与比较相当于是浪费的。第二，因为ticker本身有1秒的间隔，所以，他的超时发送ping消息的间隔并不是很准确，有可能比设置的时间还长一些。比如，我设置超时间隔为30秒，可能在未发送数据后的第31秒才会发送Ping消息。

I think, I should understand your previous suggestions, and I did. (Refer to the following websocket_wrapper.go for the changes that were not submitted before).
Just after the change, I had other ideas, which led me to make other changes. The reason for the change is this:
I mean, it's okay to do it with ticker, but there are two places that are not very satisfactory. First, the ticker will trigger every second and then perform a calculation, which will result in more useless calculations. For example, if I set the interval to 30 seconds, then the ticker performs 30 calculations, only the last calculation is valid. Except for the last calculation, other calculations and comparisons are equivalent to waste. Second, because the ticker itself has a 1 second interval, the interval between sending ping messages over time is not very accurate, and may be longer than the set time. For example, I set the timeout interval to 30 seconds, and the Ping message may be sent 31 seconds after the data is not sent.

（Translations courtesy of google translate）

`
package grpcweb

import (
	"bufio"
	"bytes"
	"context"
	"encoding/binary"
	"errors"
	"io"
	"net/http"
	"net/textproto"
	"strings"
	// "sync"
	"time"

	"github.com/gorilla/websocket"
	"golang.org/x/net/http2"
)

type webSocketResponseWriter struct {
	writtenHeaders  bool
	wsConn          *websocket.Conn
	headers         http.Header
	flushedHeaders  http.Header
	timeOutInterval time.Duration
	// lastMessageTime     time.Time
	// lastMessageTimeLock *sync.Mutex
	timer *time.Timer
}

func newWebSocketResponseWriter(wsConn *websocket.Conn) *webSocketResponseWriter {
	return &webSocketResponseWriter{
		writtenHeaders: false,
		headers:        make(http.Header),
		flushedHeaders: make(http.Header),
		wsConn:         wsConn,
	}
}

func (w *webSocketResponseWriter) EnablePing(timeOutInterval time.Duration) {
	if timeOutInterval < time.Second {
		return
	}
	w.timeOutInterval = timeOutInterval
	// w.lastMessageTimeLock = &sync.Mutex{}
	w.timer = time.NewTimer(w.timeOutInterval)
	dispose := make(chan bool)
	w.wsConn.SetCloseHandler(func(code int, text string) error {
		close(dispose)
		return nil
	})
	go w.ping(dispose)
}

func (w *webSocketResponseWriter) ping(dispose chan bool) {
	if dispose == nil {
		return
	}
	// ticker := time.NewTicker(time.Second)
	// defer ticker.Stop()
	defer w.timer.Stop()
	for {
		select {
		case <-dispose:
			return
		// case t := <-ticker.C:
		// 	var sendPing = false
		// 	w.lastMessageTimeLock.Lock()
		// 	if t.After(w.lastMessageTime.Add(w.timeOutInterval)) {
		// 		sendPing = true
		// 		w.lastMessageTime = time.Now()
		// 	}
		// 	w.lastMessageTimeLock.Unlock()
		// 	if sendPing {
		// 		w.wsConn.WriteMessage(websocket.PingMessage, []byte{})
		// 	}
		case <-w.timer.C:
			w.timer.Reset(w.timeOutInterval)
			w.wsConn.WriteMessage(websocket.PingMessage, []byte{})
		}
	}
}

func (w *webSocketResponseWriter) Write(b []byte) (int, error) {
	if !w.writtenHeaders {
		w.WriteHeader(http.StatusOK)
	}
	if w.timeOutInterval > time.Second && w.timer != nil {
		// w.lastMessageTimeLock.Lock()
		// w.lastMessageTime = time.Now()
		// w.lastMessageTimeLock.Unlock()
		w.timer.Reset(w.timeOutInterval)
	}
	return len(b), w.wsConn.WriteMessage(websocket.BinaryMessage, b)
}

`

johanbrandhorst: I'm confused. The ticker will only tick every `timeOutInterval`. It will not tick every second. With an interval of 30 seconds, what will happen is that every 30 seconds it will check if there was at least 30 seconds since the last message was sent. Sure, if you send a message 0.01s after the ping message, it will take 59.9s until the next tick is set. So worst case, you just have to configure your timeout interval to 15s if you want a message _at least_ every 30s.

angwangiot: 是的，这是我关心的。如果我是用户，我会设置一段时间的超时间隔。结果应该是ping消息发送这么长时间以满足nginx的超时要求，而不是可能是超时间隔的两倍。这需要计算以确定满足nginx要求所需的时间，并且发送Pi​​ng消息的间隔不固定，它是设置超时间隔的1到2倍。

Yes, this is what I care about. If I am a user, I will set a timeout interval for a while. The result should be that the ping message is sent for such a long time to meet the timeout requirement of nginx, rather than possibly twice the timeout interval. This requires calculations to determine the time required to meet the nginx requirements, and the interval at which the Pi ng message is sent is not fixed, it is 1 to 2 times the set timeout interval.
（Translations courtesy of google translate）

johanbrandhorst: OK, but that is simply a matter of user documentation. We can even ask for a timeout interval and set the ticker to half of that to guarantee a message between X/2 and X. I think your current implementation won't work, as you are reading from the `time.Timer` in one goroutine and resetting it in another. The documentation says you should only reset the timer if you have already read from it:
```
Reset should be invoked only on stopped or expired timers with drained channels. If a program has already received a value from t.C, the timer is known to have expired and the channel drained, so t.Reset can be used directly. If a program has not yet received a value from t.C, however, the timer must be stopped and—if Stop reports that the timer expired before being stopped—the channel explicitly drained:
```
https://golang.org/pkg/time/#Timer.Reset. I think using a ticker is the best option.

angwangiot: 好的，我明白了，感谢您耐心的解释。我先将timer改为ticker，确保此功能可以使用。
不过，我发现有个库可以规避原生timer.Reset()的限制，地址在这：https://github.com/desertbit/timer 。或者，可以考虑通过channel的方式在ping()中进行reset来保证读取和reset在同一个goroutine 。如下：
Ok, I understand, thank you for your patient explanation. I first changed the timer to ticker to make sure this feature is available.
However, I found that there is a library that can circumvent the limitations of the native timer.Reset(), where the address is: https://github.com/desertbit/timer. Or, we can consider resetting in ping() by channel to ensure that the read and reset operations are in the same goroutine.（Translations courtesy of google translate）
as follows：
`func (w *webSocketResponseWriter) ping(dispose chan bool) {
	if dispose == nil {
		return
	}
	defer w.timer.Stop()
	for {
		select {
		case <-dispose:
			return
		case <-resetTimer:
			if !w.timer.Stop() {
 				<-w.timer.C
 			}
 			w.timer.Reset(w.timeOutInterval)
		case <-w.timer.C:
			w.timer.Reset(w.timeOutInterval)
			w.wsConn.WriteMessage(websocket.PingMessage, []byte{})
		}
	}
}`
What do you think?

johanbrandhorst: Lets use the third party `timer` package, I don't like the extra complexity added by this 2nd channel implementation. Please change it and I will take another look.

go/grpcwebproxy/main.go
```diff
@@ -71,6 +72,13 @@ func main() {
 			grpcweb.WithWebsockets(true),
 			grpcweb.WithWebsocketOriginFunc(makeWebsocketOriginFunc(allowedOrigins)),
 		)
+		if *websocketPingInterval >= time.Second {
+			logrus.Println(fmt.Sprintf("websocket keepalive pinging enabled, the ping interval is %s, the actual timeout is 1-2 times the interval.", websocketPingInterval.String()))
```
johanbrandhorst: You can use `logrus.Infof` to get string interpolation.

go/grpcweb/options.go
```diff
@@ -92,6 +96,15 @@ func WithWebsockets(enableWebsockets bool) Option {
 	}
 }
 
+// WithWebsocketPingInterval enables websocket keepalive pinging with the configured timeout (in seconds).
```
johanbrandhorst: This is a `time.Duration` now so we don't need to specify the unit anymore.
```suggestion
// WithWebsocketPingInterval enables websocket keepalive pinging with the configured timeout.
```

angwangiot: 十分感谢您一直以来的帮助。昨天，在提交之前我已经在我的环境中测试通过了。
1.在未使用websocket keepalive pinging时，超时后没有Ping消息，并且60s后就断开了连接。
图片
2.在将websocket_ping_interval设置为20s时，每次超时20s，会发送Ping消息，并且超时时间比较精确，同时，nginx不在断开连接。
图片
这与预期的结果相吻合。

Thank you very much for your continuous help.Yesterday, I tested it in my environment before submitting it.
1.When the websocket keepalive pinging is not used, there is no Ping message after timeout, and the connection is disconnected after 60s.
2.When the websocket_ping_interval is set to 20s, a Ping message is sent with each 20s timeout, and the timeout time is accurate, and nginx is no longer disconnected.
This is in line with the expected result. （Translations courtesy of google translate）

johanbrandhorst:
Thanks for this contribution @angwangiot, great work!

angwangiot:
It's my pleasure to participate in development and to be able to contribute.

PR #557, opened 2019-09-25
Migrate to go modules

johanbrandhorst:
This is way overdue already. Lets see if CI passes.


johanbrandhorst:
Hm, adding `1.12` and `1.13` has tripled the number of integration tests. @jonny-improbable can we drop `1.11` and `1.12` and just run `1.13`?

jonny-improbable:
Works for me :)

devnev:
Regarding go versions, the last three major versions are considered supported by the language developers themselves: https://golang.org/doc/devel/release.html#policy
Also, supporting at least two versions would ensure users always have a safe upgrade path (in this case the two versions being the 1.11 that was being supported previously and the 1.13 that will be supported in the future).

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: go
 go:
-  - "1.11"
+  - "1.13"
```
devnev: From my understanding of https://github.com/travis-ci/gimme#asking-gimme-what-a-version-is, specifying `1.13.x` would automatically update to newer patch versions, would it make sense to use that?

johanbrandhorst: Good idea!

johanbrandhorst:
> Regarding go versions, the last three major versions are considered supported by the language developers themselves: https://golang.org/doc/devel/release.html#policy

This is a common misunderstanding - reread the document again, they support the last _two_ major versions.

> Also, supporting at least two versions would ensure users always have a safe upgrade path (in this case the two versions being the 1.11 that was being supported previously and the 1.13 that will be supported in the future).

This makes sense, but given what I said above I think it's okay to just support 1.12 and 1.13. The problem here is that it adds to the already slow CI system. I'd be happy to _say_ we support 1.12 but not actually test for it. I don't have any reason to expect that we don't.

johanbrandhorst:
Think it needs a bit of work still so I stopped the CI jobs. Looks like the intergration tests expects the go server to start immediately.

johanbrandhorst:
Actually the Go 1.12 tests were doing okay.

johanbrandhorst:
@jonny-improbable ok I have no idea whats wrong with the integration tests. Seem to work on 1.12 and not 1.13. Node failure logs couldn't be less helpful. Could you try to clone this and see if you can get it running locally? I don't think there should be any go issues.

johanbrandhorst:
https://travis-ci.org/improbable-eng/grpc-web/jobs/589481073 here's a very unhelpful log.

jonny-improbable:
> @jonny-improbable <https://github.com/jonny-improbable> ok I have no idea
whats wrong with the integration tests. Seem to work on 1.12 and not 1.13.
Node failure logs couldn't be less helpful. Could you try to clone this and
see if you can get it running locally? I don't think there should be any go
issues.

Will try and put some time into this tomorrow, but am pretty slammed at
present - let me see what I can do.

johanbrandhorst:
Superceded by #911

PR #576, opened 2019-10-11
Removed detach

MarcusLongmuir:
This PR removes the `detach` function and selectively replaces its usage.

The `detach` function was used to isolate user-provided callbacks from the rest of the `grpc-web` such that exceptions thrown by user-provided callbacks would not be caught by `grpc-web`. It called the user-provided callbacks inside a `setTimeout` call to prevent the catching of exceptions from the callback function. In my own usage of `grpc-web` with fake transports and alike this proved frustrating due to the disconnected stack trace caused by the `setTimeout`.

The function was also overused as a protective measure inside transports and that usage has been removed completely.

The usages that wrapped user-provided callbacks have been replaced by a `try+catch` that uses `setTimeout` to re-throw the exception if the callback throws one, but otherwise will execute the callback immediately. 

**This is a breaking change**.

Usage that assumed the callbacks provided to `grpc-web` would be executed in a deferred manner could be broken by this change. This is unlikely to affect production code as the network request will always be asynchronous and would cause the callback to be invoked sometime in the future, but tests that assume other code will run between the `grpc-web` invocation and the callback being called could be affected.


jonny-improbable:
@MarcusLongmuir what is the status of this PR? I recall you mentioning it may improve performance, is that still the case?

MarcusLongmuir:
@jonny-improbable: I've just managed to get this passing tests and it's now ready for review.

client/grpc-web-fake-transport/src/index.spec.ts
```diff
@@ -233,6 +246,7 @@ describe("FakeTransportBuilder", () => {
       const onEndSpy = jest.fn();
 
       const transport = new FakeTransportBuilder()
+        .withManualTrigger()
```
jonny-improbable: Duplicated line.

client/grpc-web-fake-transport/src/index.spec.ts
```diff
@@ -20,6 +21,7 @@ describe("FakeTransportBuilder", () => {
     it("should allow the response trailers to be stubbed", done => {
       const expectedTrailers = new grpc.Metadata({ foo: "bar" });
       const transport = new FakeTransportBuilder()
+        .withManualTrigger()
```
jonny-improbable: Hmm, I don't understand what's going on here - could you please explain this change? I see you've made it be a manual trigger, but I don't see you triggering it anywhere in the test?

client/grpc-web-fake-transport/src/index.spec.ts
```diff
@@ -258,8 +272,6 @@ describe("FakeTransportBuilder", () => {
 
       transport.sendHeaders();
 
-      // Note that we must defer the assertions as the grpc-web package detaches all callbacks, so they
-      // will not run immediately.
       setTimeout(() => {
```
jonny-improbable: You've removed the comment, but not the timeout :)

galenwarren:
Hi guys -- first, thanks for your work on this library, it's great!

I was having an issue where I would occasionally get errors logged in the console associated with connections that were being closed and which were also receiving data from a (server) streaming grpc call at the same time. It doesn't happen all the time but does happen frequently when those conditions are met, maybe a race condition of some sort? Not sure.

Here's an example of the error logged in the browser console:

>VM11014:1 POST https://realurlreplaced net::ERR_FAILED 200
(anonymous)	@	VM11014:1
e.send	@	grpc-web-client.umd.js?ba3e:1
e.sendMessage	@	grpc-web-client.umd.js?ba3e:1
e.send	@	grpc-web-client.umd.js?ba3e:1
t.invoke	@	grpc-web-client.umd.js?ba3e:1

To try to figure out what was going on, I cloned the repo. Using the code in the master branch, I can't reproduce the error, whereas I can reproduce it consistently with the 0.12.0 code.

Looking at the commits in master since 0.12.0, [this](https://github.com/improbable-eng/grpc-web/compare/v0.12.0...master#diff-5b66a06591e916a25ed5fa86c3da981e) looks like the potentially relevant change.

So, I wanted to let you know this latest code seems to be addressing my issue -- thanks! -- and also to ask whether you're considering releasing this code anytime soon?

MarcusLongmuir:
@galenwarren: We've just released `0.13.0` from latest `master` so you should be good to update to that to resolve the issue. Thanks for the prompt as it had been missed that this change hadn't gone out 👍

galenwarren:
Awesome -- thanks!

Issue #564, opened 2019-10-01
grpcwebproxy flag parsing is a bit broken


Timmmm:
Ok I found out what was causing my weird issues in #562 and #563 - the flag parsing for bool parameters is kind of broken. Basically this:

     ./grpcwebprox --run_tls_server false

Means it *will* run the tls server. It doesn't produce an error or anything. You need to do this:

     ./grpcwebprox --run_tls_server=false

Ok crazy. Maybe you just always need to use `=`? No you don't! `--backend_addr 127.0.0.1` works fine. It is just the boolean flags that have this weird broken parsing.

johanbrandhorst:
Yeah, this is a bit weird, I can see the reason for the confusion. However, this is how POSIX/GNU style flags work, apparently according to the library we're using: https://github.com/spf13/pflag#command-line-flag-syntax. I don't think we're going to change library, so while I'm sorry this caused some confusion for you, I don't think we're going to do anything about it.

Timmmm:
It says:

> -b true is INVALID

That implies that it should give some kind of error. It also says:

> flags can be interspersed with arguments anywhere on the command line

So I think it is treating `true` as an argument (not a flag). In that case you should be processing arguments, and since the command doesn't support any arguments, give an error if there are any.

I think you can do it something like this (not tested at all):

```
func main() {
	pflag.Parse()
	if len(pflag.Args()) != 0 {
		for _, flag := range pflag.Args() {
			if flag == "true" || flag == "false" {
				logrus.Fatal("Boolean flags should be set using `--flag=false`, `--flag=true` or `--flag` (which is short for `--flag=true`). You cannot use `--flag true` or `--flag false`.")
			}
		}
		logrus.Fatal("Unknown argument: " + flag)
	}
	...
```

johanbrandhorst:
Sounds worthwhile, since you ran into this issue, would you be interested in contributing this fix?

PR #577, opened 2019-10-15
Check that there aren't any unknown command line arguments

Timmmm:
See #564

## Changes

* Check that there are no unknown command line arguments.

## Verification

I haven't been able to test this because I get  a compilation error when building:

```
# github.com/grpc-ecosystem/go-grpc-prometheus
/Users/me/go/src/github.com/grpc-ecosystem/go-grpc-prometheus/client_reporter.go:51:29: cannot use prometheus.NewTimer(hist) (type *prometheus.Timer) as type timer in return argument:
        *prometheus.Timer does not implement timer (wrong type for ObserveDuration method)
                have ObserveDuration()
                want ObserveDuration() time.Duration
/Users/me/go/src/github.com/grpc-ecosystem/go-grpc-prometheus/client_reporter.go:64:29: cannot use prometheus.NewTimer(hist) (type *prometheus.Timer) as type timer in return argument:
        *prometheus.Timer does not implement timer (wrong type for ObserveDuration method)
                have ObserveDuration()
                want ObserveDuration() time.Duration
```

### DIFF
--- a/go/grpcwebproxy/main.go
+++ b/go/grpcwebproxy/main.go
@@ -48,6 +48,14 @@ var (
 
 func main() {
 	pflag.Parse()
+	if len(pflag.Args()) != 0 {
+		for _, flag := range pflag.Args() {
+			if flag == "true" || flag == "false" {
+				logrus.Fatal("Boolean flags should be set using --flag=false, --flag=true or --flag (which is short for --flag=true). You cannot use --flag true or --flag false.")
+			}
+		}
+		logrus.Fatal("Unknown argument: " + flag)
+	}
 
 	logrus.SetOutput(os.Stdout)
 	logEntry := logrus.NewEntry(logrus.StandardLogger())
--- a/go/grpcwebproxy/main.go
+++ b/go/grpcwebproxy/main.go
@@ -48,11 +48,9 @@ var (
 
 func main() {
 	pflag.Parse()
-	if len(pflag.Args()) != 0 {
-		for _, flag := range pflag.Args() {
-			if flag == "true" || flag == "false" {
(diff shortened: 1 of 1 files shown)


go/grpcwebproxy/main.go
```diff
@@ -48,6 +48,14 @@ var (
 
 func main() {
 	pflag.Parse()
+	if len(pflag.Args()) != 0 {
```
johanbrandhorst: This check is unnecessary, there will just not be a loop if it's 0.

go/grpcwebproxy/main.go
```diff
@@ -48,6 +48,14 @@ var (
...
+		for _, flag := range pflag.Args() {
+			if flag == "true" || flag == "false" {
+				logrus.Fatal("Boolean flags should be set using --flag=false, --flag=true or --flag (which is short for --flag=true). You cannot use --flag true or --flag false.")
+			}
+		}
+		logrus.Fatal("Unknown argument: " + flag)
```
johanbrandhorst: What is this line for? It is outside the loop and it's referencing a variable that doesn't look like it's defined.

Timmmm:
Oops! Fixed broken code, and it builds correctly with that `dep` command. Output is:

```
$ ./grpcwebproxy --run_http_server false
FATA[0000] Boolean flags should be set using --flag=false, --flag=true or --flag (which is short for --flag=true). You cannot use --flag true or --flag false.
$ ./grpcwebproxy --run_http_server=false
FATA[0000] flag 'backend_addr' must be set
$ ./grpcwebproxy --run_http_server sfgdfg
FATA[0000] Unknown argument: sfgdfg
```

johanbrandhorst:
Thanks for your contribution!

Timmmm:
Thanks for merging quickly! :-)

Issue #197, opened 2018-05-27
Support for Bazel


igorgatis:
It would be great to have rules for Bazel, specially if they work with https://github.com/bazelbuild/rules_typescript

easyCZ:
Hey @igorgatis,

Indeed. Currently working on https://github.com/improbable-eng/ts-protoc-gen/issues/77. Once we have that, grpc-web is the natural next step.

Having said that, your help would be greatly appreciated. If you've got experience with Bazel or want to dig in, please do. We can use this channel to figure out the details.

mikeleonard:
Hey @easyCZ. Just came across this issue looking for bazel support for grpc-web.

Just FYI I noticed that this pull request https://github.com/improbable-eng/ts-protoc-gen/pull/84 has been merged and so the ts-protoc-gen bazel rule is I think complete but issue https://github.com/improbable-eng/ts-protoc-gen/issues/77 hasn't been closed yet. 

Thought I'd just flag it here as you're waiting for https://github.com/improbable-eng/ts-protoc-gen/issues/77 to move forward with bazel support for grpc-web - which I selfishly want :)

coltonmorris:
Isn't there already bazel support? It would be nice if rules-typescript allowed ts code to import the files generated by ts-protoc-gen in a cleaner way rather than having to reference gen files or use path aliases. But it's still supported and works.

Maybe I'm missing something? ¯\_(ツ)_/¯

johanbrandhorst:
Do we need both `Bazel` support and `go.mod` support? Can Bazel accommodate go mod files? See #202

PR #627, opened 2019-12-26
Provide bazel BUILD files for grpcweb and grpcwebproxy.

dprogm:
Added bazel BUILD files in order to be able building `grpcweb` and `grpcwebproxy` with the bazel build system. I already use grpcwebproxy and build it with bazel so I thought maybe we can add it here too. One thing to notice: I couldn't compile the tests for grpcweb. I have actually almost no knowledge about go. Maybe someone can help me to get it compile and run.

Not sure, but it could solve parts of https://github.com/improbable-eng/grpc-web/issues/197

## Changes

* Bazel WORKSPACE setup
* Added all required go dependencies
* Added build targets for grpcweb and grpcwebproxy

## Verification

```
bazel build ...
bazel run //go/grpcwebproxy -- --help
```

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,0 +1,1 @@
+build --symlink_prefix=dist/
--- a/.gitignore
+++ b/.gitignore
@@ -168,3 +168,6 @@ misc/localhostCA.srl
 
 node_modules
 vendor
+
+bazel-out
+.DS_Store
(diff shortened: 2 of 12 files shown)


johanbrandhorst:
Hi, thanks for your interest in this project! Personally I'm not sure we _want_ to add these, as our CI doesn't use bazel, but I think Improbable might use bazel internally so I'll leave the decision up to the maintainers who are also employees at Improbable. Regardless, thanks for your contribution!

PR #660, opened 2020-02-21
fix example

soyersoyer:
## Changes
it fixes the grpcweb example


johanbrandhorst:
Ah, the CI failed because we generate this. Can you run `./generate-docs.sh`?

soyersoyer:
sure!

PR #670, opened 2020-03-13
Make example code a little friendlier

achew22:
Previously the example code didn't check if the request was a CORS which meant that the `OPTIONS` preamble for the CORS request would be handled by the `DefaultServeMux.ServeHTTP` method which probably isn't correct. Additionally you should probably only have one handler that handles every request.


johanbrandhorst:
Apparently the CI failure is because we generate the docs, could you have a look at the error? It could be we generate this from an actual file, or it could just be some spaces are in the wrong place.

PR #736, opened 2020-07-20
Docs: Fixes a deprecated link for Alt. Transports

sam-parsons:
## Changes

Current link points to a deprecated, un-prefixed version of the package, corrected to most recent version.


johanbrandhorst:
@MarcusLongmuir @jonny-improbable we really need to do something about these tests. I'm leaning towards just turning them off when they're this unreliable. What do you think?

johanbrandhorst:
They failed again on the rerun. Something is up.

MarcusLongmuir:
I'll investigate now

Issue #740, opened 2020-07-21
Secured CI prevents external contributions from being tested


MarcusLongmuir:
First identified on #736 

Due to changes made in #718 the browserstack parameters are now secured using Travis CI `secure` parameters:
https://docs.travis-ci.com/user/encryption-keys/#usage

However, this creates an understandable limitation of not allowing including the values of these secured parameters when a PR is opened from a fork as this would expose the values to potentially malicious code:
https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions

An ideal solution in my opinion would be that the browserstack-based tests in the Travis CI run are blocked until a maintainer comments to that effect (based on checking the code for malicious intent) as I have seen other open source projects do.

cc @johanbrandhorst

johanbrandhorst:
Sounds good to me, how could we set up such a trigger?

PR #760, opened 2020-08-25
grpcweb: Websockets test should use case-insensitive header compare

mgkeeley:
When checking if a request is a websocket request, the `Upgrade` header is checked to see if it contains the text `websocket`. However, IE11 sends the text `Websocket` (capital W).  The comparison should be made case-insensitive to support browsers that send non-standard requests.
Yes, I know IE11 is going away, but this browser is still used by the WinForms WebBrowser control, for those poor souls that must continue deal with it.

## Changes

Convert the header values to lower case before comparing in `IsGrpcWebSocketRequest`.

## Verification

Tried a small example manually in IE11 and latest Chrome (Win7 64bit).
I couldn't get the tests to run locally.

### DIFF
--- a/go/grpcweb/wrapper.go
+++ b/go/grpcweb/wrapper.go
@@ -133,7 +133,7 @@ func (w *WrappedGrpcServer) ServeHTTP(resp http.ResponseWriter, req *http.Reques
 // IsGrpcWebSocketRequest determines if a request is a gRPC-Web request by checking that the "Sec-Websocket-Protocol"
 // header value is "grpc-websockets"
 func (w *WrappedGrpcServer) IsGrpcWebSocketRequest(req *http.Request) bool {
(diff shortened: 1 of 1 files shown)


johanbrandhorst (approved):
SGTM, sorry for the inconvenience.

mgkeeley:
thanks :+1:

johanbrandhorst:
@jonny-improbable @MarcusLongmuir Is prow supposed to be running on this repo? We don't have an OWNERS

johanbrandhorst:
CI failed :cry:. Doesn't look like it's got anything to do with the change though. @MarcusLongmuir?

MarcusLongmuir:
As a workaround for #740 I pushed this branch to this repo and ran it manually: https://travis-ci.org/github/improbable-eng/grpc-web/builds/721294693 ✅

PR #588, opened 2019-10-28
Remove connection header from grpcwebproxy

crlssn:
This PR fixes the issue expressed in https://github.com/improbable-eng/grpc-web/issues/568. Basically, when you are sending a gRPC request from a web client through the proxy, multiple headers are added of which `connection` is denying the actual connection to be established.

Error message with the `connection` header present:
```
INFO[0000] listening for http on: [::]:8080             
INFO[0000] pickfirstBalancer: HandleSubConnStateChange: 0xc00015e200, READY  system=system
INFO[0075] transport: loopyWriter.run returning. connection error: desc = "transport is closing"  system=system
INFO[0075] pickfirstBalancer: HandleSubConnStateChange: 0xc00015e200, TRANSIENT_FAILURE  system=system
INFO[0075] pickfirstBalancer: HandleSubConnStateChange: 0xc00015e200, CONNECTING  system=system
WARN[0075] finished streaming call with code Unavailable  error="rpc error: code = Unavailable desc = transport is closing" grpc.code=Unavailable grpc.method=Method grpc.service=namespace.Service grpc.start_time="2019-10-28T11:56:35Z" grpc.time_ms=1052.424 span.kind=server system=grpc
```

PR #795, opened 2020-11-01
Drop proxy-connection header

sucy6330133:
## Changes

Drop the 'Proxy-Connection' header from the grpcwebproxy.
This is related to issue #588, when testing ui with Cypress, instead adding the connection header, the {'proxy-connection': 'keep-alive'} header has been added by Cypress. 
This is the error log being printed out by grpcwebproxy
```
ERRO[0215] finished streaming call with code Internal    error="rpc error: code = Internal 
desc = stream terminated by RST_STREAM with error code: PROTOCOL_ERROR" 
grpc.code=Internal grpc.method=Echo grpc.service=grpc.gateway.testing.EchoService grpc.start_time="2020-11-02T08:23:43+13:00" grpc.time_ms=0.345 span.kind=server system=grpc
```
## Verification

By comparing the requests sent through Chrome and Cypress, the only difference is the connection header and proxy-connection header. After drop the proxy-connection header, the grpcwebproxy forwarded the request to node server successfully and the cypress tests passed.
/assign

### DIFF
--- a/go/grpcwebproxy/main.go
+++ b/go/grpcwebproxy/main.go
@@ -156,6 +156,7 @@ func buildGrpcProxyServer(logger *logrus.Entry) *grpc.Server {
 		// the actual connection to the backend will not be established.
 		// https://github.com/improbable-eng/grpc-web/issues/568
 		delete(mdCopy, "connection")
+		delete(mdCopy, "proxy-connection")
 		outCtx = metadata.NewOutgoingContext(outCtx, mdCopy)
 		return outCtx, backendConn, nil
 	}


johanbrandhorst:
That's super weird. Before merging something like this, could you add a test that fails without this fix? It feels a bit strange that we should add this here if it's a consequence of the environment in which it is used. I'd rather users drop this header manually in another proxy sitting in front of this one. If you can write a test that fails before this fix is applied then I'd be more convinced of its legitimacy.

sucy6330133:
Which test you specified here? Unit tests?

johanbrandhorst:
No, ideally a new integration test (we have an existing structure [here](https://github.com/improbable-eng/grpc-web/tree/master/integration_test/ts)). If this header changes the behaviour of the proxy as your theory states, it should be possible to create a failing test that mimics the circumstances of your situation that requires this change.

sucy6330133:
Is the integration test structure working? I have followed the [instructions](https://github.com/improbable-eng/grpc-web/blob/master/CONTRIBUTING.md) and have issue to run the tests locally. 
I firstly tried the command 'npm run test:integration', and got error:
```
2020-11-05T12:01:01.474+1300	WARN	protoc returned a line we do not understand, please file this as an issue at https://github.com/uber/prototool/issues/new	{"protocLine": "2020/11/05 12:01:01 WARNING: Missing 'go_package' option in \"improbable/grpcweb/test/test.proto\","}
2020-11-05T12:01:01.474+1300	WARN	protoc returned a line we do not understand, please file this as an issue at https://github.com/uber/prototool/issues/new	{"protocLine": "please specify it with the full Go package path as"}
2020-11-05T12:01:01.474+1300	WARN	protoc returned a line we do not understand, please file this as an issue at https://github.com/uber/prototool/issues/new	{"protocLine": "a future release of protoc-gen-go will require this be specified."}
2020-11-05T12:01:01.474+1300	WARN	protoc returned a line we do not understand, please file this as an issue at https://github.com/uber/prototool/issues/new	{"protocLine": "See https://developers.google.com/protocol-buffers/docs/reference/go-generated#package for more information."}
<input>:1:1:2020/11/05 12:01:01 WARNING: Missing 'go_package' option in "improbable/grpcweb/test/test.proto",
<input>:1:1:See https://developers.google.com/protocol-buffers/docs/reference/go-generated#package for more information.
<input>:1:1:a future release of protoc-gen-go will require this be specified.
<input>:1:1:please specify it with the full Go package path as
<input>:1:1:protoc-gen-go: protoc-gen-go: plugins are not supported; use 'protoc --go-grpc_out=...' to generate gRPC
```

sucy6330133:
Anyway, I have found this 
image
The grpc is HTTP/2-based transport, I think the 'Proxy-Connection' header needs to be dropped.

johanbrandhorst:
> Is the integration test structure working? I have followed the [instructions](https://github.com/improbable-eng/grpc-web/blob/master/CONTRIBUTING.md) and have issue to run the tests locally.
> I firstly tried the command 'npm run test:integration', and got error:
> 
> ```
> 2020-11-05T12:01:01.474+1300	WARN	protoc returned a line we do not understand, please file this as an issue at https://github.com/uber/prototool/issues/new	{"protocLine": "2020/11/05 12:01:01 WARNING: Missing 'go_package' option in \"improbable/grpcweb/test/test.proto\","}
> 2020-11-05T12:01:01.474+1300	WARN	protoc returned a line we do not understand, please file this as an issue at https://github.com/uber/prototool/issues/new	{"protocLine": "please specify it with the full Go package path as"}
> 2020-11-05T12:01:01.474+1300	WARN	protoc returned a line we do not understand, please file this as an issue at https://github.com/uber/prototool/issues/new	{"protocLine": "a future release of protoc-gen-go will require this be specified."}
> 2020-11-05T12:01:01.474+1300	WARN	protoc returned a line we do not understand, please file this as an issue at https://github.com/uber/prototool/issues/new	{"protocLine": "See https://developers.google.com/protocol-buffers/docs/reference/go-generated#package for more information."}
> <input>:1:1:2020/11/05 12:01:01 WARNING: Missing 'go_package' option in "improbable/grpcweb/test/test.proto",
> <input>:1:1:See https://developers.google.com/protocol-buffers/docs/reference/go-generated#package for more information.
> <input>:1:1:a future release of protoc-gen-go will require this be specified.
> <input>:1:1:please specify it with the full Go package path as
> <input>:1:1:protoc-gen-go: protoc-gen-go: plugins are not supported; use 'protoc --go-grpc_out=...' to generate gRPC
> ```

These are just warnings, I wouldn't worry about them. We do need to update our go generator eventually but it won't cause any test failures. Please do try to add a test, and if we need to update the docs lets do that too.

>  The grpc is HTTP/2-based transport, I think the 'Proxy-Connection' header needs to be dropped.

This is really compelling, what is the source for this? Could you provide a link?

I'd still like to see a failing test, which should not be too hard (just copy one of the echo tests and set the Proxy-Connection header and watch it fail).

sucy6330133:
> This is really compelling, what is the source for this? Could you provide a link?
> 
https://en.wikipedia.org/wiki/List_of_HTTP_header_fields

sucy6330133:
> I'd still like to see a failing test, which should not be too hard (just copy one of the echo tests and set the Proxy-Connection header and watch it fail).
I didn't find anything existing tests which is verified the grpcwebproxy. Would you mind pointing it out?

johanbrandhorst:
[Here](https://github.com/improbable-eng/grpc-web/blob/master/integration_test/ts/src/client.spec.ts#L86-L128) is a test that makes a unary request to the backend. You could copy paste this test and try to add the proxy-connection header (I don't know exactly how to do this in the client, it might not be trivially exposed).

sucy6330133:
> [Here](https://github.com/improbable-eng/grpc-web/blob/master/integration_test/ts/src/client.spec.ts#L86-L128) is a test that makes a unary request to the backend. You could copy paste this test and try to add the proxy-connection header (I don't know exactly how to do this in the client, it might not be trivially exposed).

Modify the test here to send the 'proxy-header' to server will not verify the grpcwebproxy can handle the 'proxy-connection' header, right?

johanbrandhorst:
I don't see why it wouldnt? I'm basically asking you to reproduce the environment you're experiencing the issue in so that we have a test that can verify that we fixed it.

sucy6330133:
I wrote the integration tests and the server is happy with the 'Proxy-Connection' header. Then I have retried the grpcwebproxy again to use the wireshark to see the packets. I don't know what I did wrong previously, but it worked this time.
I guess the pull request is no longer needed and I am sorry to cause any confusion.

MarcSeebold:
Please re-open this MR. Our cypress instance adds the `proxy-connection` header and we can't turn that off. I applied this patch and it solved that issue. Furthermore, I don't see any risks with this patch since `proxy-connection` is not supported by HTTP/2 whatsoever.

PR #811, opened 2020-12-11
Refactor websocket handler to protect against concurrent writes

### LINKED ISSUE #713: grpcwebproxy: "panic: concurrent write to websocket connection"
I get the following error when running the grpxwebproxy service:

```
panic: concurrent write to websocket connection

goroutine 7407 [running]:
github.com/improbable-eng/grpc-web/vendor/github.com/gorilla/websocket.(*messageWriter).flushFrame(0xc00060beb0, 0xeec501, 0xeec5e0, 0x0, 0x0, 0x13da78305d1bc0, 0xc00060beb8)
	/Users/jonnyreeves/Projects/grpc-web/src/github.com/improbable-eng/grpc-web/vendor/github.com/gorilla/websocket/conn.go:591 +0x6f6
github.com/improbable-eng/grpc-web/vendor/github.com/gorilla/websocket.(*Conn).WriteMessage(0xc0004a8dc0, 0x9, 0xeec5e0, 0x0, 0x0, 0x0, 0x0)
	/Users/jonnyreeves/Projects/grpc-web/src/github.com/improbable-eng/grpc-web/vendor/github.com/gorilla/websocket/conn.go:752 +0x24c
github.com/improbable-eng/grpc-web/go/grpcweb.(*webSocketResponseWriter).ping(0xc000387920, 0xc0000cb800)
	/Users/jonnyreeves/Projects/grpc-web/src/github.com/improbable-eng/grpc-web/go/grpcweb/websocket_wrapper.go:60 +0xbd
created by github.com/improbable-eng/grpc-web/go/grpcweb.(*webSocketResponseWriter).enablePing
	/Users/jonnyreeves/Projects/grpc-web/src/github.com/improbable-eng/grpc-web/go/grpcweb/websocket_wrapper.go:46 +0xea
```

grpcwebproxy-v0.12.0-linux-x86_64
grpcwebproxy-v0.12.0-win64.exe

**What happened**
Very intermittently, the proxy process crashes with the message 
```panic: concurrent write to websocket connection```

No crash :-)

I can't reproduce this reliably, it is intermittent on our system.

According to https://godoc.org/github.com/gorilla/websocket#hdr-Concurrency this is because there must only be one writer at a time.

My guess is that the ping happened at exactly the same time as some other message.

We use both linux and windows pre-built versions of the proxies, I saw this error on linux.

### PR DESCRIPTION
**Pull request by pcj**
1. Move websocket http handler code from `wrapper.go` to `websocket_handler.go`.
2. Added `websocket_client.go` and `websocket_clienthub.go` to manage concurrent requests.
3. Broke out the reader / writer structs into separate files.
3. Refactored `webSocketResponseWriter` to send messages to the client hub rather than write to the websocket directly.

Still working on this.  Needs a proper test.

Closes #713

### CODE REVIEW

File: go/grpcweb/websocket_writer.go [hunk 1]
```diff
@@ -0,0 +1,101 @@
...
+		// See "HTTP wire protocols" section in
+		// https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md#protocol-differences-vs-grpc-over-http2
+		keyCase(strings.ToLower),
+	)
+	return th
+}
+
+func (w *webSocketResponseWriter) writeMessage(b []byte) error {
+	w.client.send <- b
+	return nil // TODO(pcj): is there a need to propagate write errors?
```
johanbrandhorst, line 91: I think at best we'll end up logging them, so we probably can just log it in the websocket writer.

go/grpcweb/websocket_writer.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func (w *webSocketResponseWriter) flushTrailers() {
+	w.writeHeaderFrame(w.extractTrailerHeaders())
+}
+
+// Flush implements the https://golang.org/pkg/net/http/#Flusher interface.
+func (w *webSocketResponseWriter) Flush() {
+	// no-op
+}
```
johanbrandhorst, line 101: Was this always a no-op, or can we coerce the websocket writer to implement this?

johanbrandhorst (commented):
Nice work, really liking where this is going. I'm wondering, if we're using significant amounts of the code from the gorilla library, should we include a copy of their license in this directory too? Their license isn't in conflict with ours, I hope?

johanbrandhorst:
FYI https://github.com/improbable-eng/grpc-web/pull/815 was just opened as an alternative fix for the underlying issue.

pcj:
#815 looks like a simpler change.  I'll test that one out in my environment and if it seems OK, will plan to close this PR.  I did make some significant inroads on a test suite for the websocket functionality; I can contribute that as a separate PR.

johanbrandhorst:
Thanks a lot, Paul 🙏🏻, that sounds good.

Issue #480, opened 2019-06-13
Flush() being called for ResponseWriter which doesn't implement the interface


mangas:
**Versions of relevant software used**
v0.9.6

**What happened**
Panic when cast to interface

**What you expected to happen**
Method should only be called upon successful cast

**How to reproduce it (as minimally and precisely as possible)**:
Call webGrpcResponse.Flush() with `wrapped` being an instance of http.ResponseWriter

mangas:
There are 2 reason why I think this should be solved in the wrapper. 

1) The api semantics don't specify anything regarding the `ResponseWriter` having to implement `http.Flusher`and should in that case treat as an optional feature or change the api to take a `Flusher` instance. 

2) Implementing a "dummy" `Flush` func also means flush might be called assuming the normal semantics and in fact nothing will happen

PR #817, opened 2020-12-16
Check flusher before flush

kostyay:
## Changes
Fixes https://github.com/improbable-eng/grpc-web/issues/480

Better implementation of https://github.com/improbable-eng/grpc-web/pull/479 without the issue it introduced

## Verification

Behaviour is unchanged for the standard case


johanbrandhorst (approved):

johanbrandhorst:
Not sure what's going on with the tests, but this should not change any behaviour so I might override the tests here unless there are any objections.

johanbrandhorst:
I gave this two days, merging now.

kostyay:
Thanks :)

PR #822, opened 2020-12-28
websocket: expose WebSocketWrapped for only wrap websocket

Junnplus:
I use grpc-web and grpc-gateway at the same time for handler different request,
and I also need handler websocket to grpc-gateway, 
I hope grpcweb module can expose WebSocketWrapped for only wrap websocket.

## Changes

1. make WebSocketWrapped inherit from *WrappedGrpcServer
2. moved websocket related from wrapper.go to websocket_wrapper.go as method of WebSocketWrapped

### DIFF
--- a/go/grpcweb/DOC.md
+++ b/go/grpcweb/DOC.md
@@ -175,6 +175,54 @@ bidirectional requests.
 
 The default behaviour is false, i.e. to disallow websockets
 
+#### type WebSocketWrapped
+
+```go
+type WebSocketWrapped struct {
+	*WrappedGrpcServer
+}
+```
+
+
+#### func  WrapHandlerWithOnlyWebsocket
+
+```go
+func WrapHandlerWithOnlyWebsocket(handler http.Handler, options ...Option) *WebSocketWrapped
+```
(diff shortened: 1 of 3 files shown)


johanbrandhorst:
Hi @Junnplus, thanks for the PR. I'm confused, what exactly is the use case that necessitates this change? It should be no problem running the grpc-web proxy and grpc-gateway alongside each other already. I'm hesitant to add support for this unless I understand there's a usecase that isn't otherwise supported.

Junnplus:
@johanbrandhorst thanks for your reply. I use grpc-web for handler websocket request from non-browser,
I also need handler websocket -> grpc-gateway from browser, I hope can reuse websocket implementation of grpc-web

johanbrandhorst:
I see, well, I don't think that's necessarily something we want to add support for. I encourage you to do any grpc-gateway routing before you perform the websocket upgrade and hand over the handling to the grpc-web websocket handler. Adding this functionality would increase the complexity of our code, which is already more than we can manage. Sorry for the inconvenience.

I'm going to close this, feel free to fork the proxy if you want to add this functionality yourself.

PR #837, opened 2021-01-20
Dependency updates

kmilner:
This updates several dependencies through a combination of `lerna-audit` and `npm audit`.  For the moment, this is only safe dependency updates, so this should not cause any problems.

Once this is merged I will review the remaining dependabot issues which will need more manual intervention.


MarcusLongmuir (approved):
LGTM. Thanks for resolving these.

I've got a branch from a while ago that I was intending to update all of the dependencies on as there are quite a few config changes from some of the major bumps IIRC.

That effort evolved into, and was then separated from, the CI update that landed a few weeks ago so I need to go back and rebase that update branch.

MarcusLongmuir (approved):
LGTM. Thanks for resolving these.

I've got a branch from a while ago that I was intending to update all of the dependencies on as there are quite a few config changes from some of the major bumps IIRC.

That effort evolved into, and was then separated from, the CI update that landed a few weeks ago so I need to go back and rebase that update branch.

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution!

PR #841, opened 2021-01-30
Update TS/JS & Go dependencies

MarcusLongmuir:
## Changes

* Updated TS/JS dependencies using `ncu -u`
* Updated Golang dep dependencies

## Verification

* Manual testing of the testrunner server, local integration test run, and client example project
* Will observe CI run


Gopkg.toml
```diff
@@ -32,29 +32,24 @@ required = [
...
 
 [[constraint]]
   name = "google.golang.org/grpc"
-  version = "1.8.0"
-
-# To work around github.com/grpc-ecosystem/go-grpc-middleware requiring master
-[[override]]
-  name = "github.com/golang/protobuf"
-  version = "1.3.0"
+  version = "1.35.0"
```
johanbrandhorst, line 55: FYI this version of gRPC depends on the APIv2 of Go protobufs: https://github.com/grpc/grpc-go/blob/v1.35.x/go.mod#L16. We may want to lock to [1.32.x](https://github.com/grpc/grpc-go/blob/v1.32.x/go.mod#L9) which seems to be the latest major using APIv2.

johanbrandhorst (approved):
Small comment. Can I just state how nice it is to have reliable CI? Thanks @MarcusLongmuir! Next steps:

- Consider migrating Go protobufs to APIv2. This may be blocked on parts of the Improbable infrastructure migrating too.
- (Cheeky suggestion) migrate protobuf generation and linting from `prototool` to `buf`. Don't take my word for it, the maintainers of `prototool` suggest switching to `buf` too: https://github.com/uber/prototool#prototool. There is a page on our doc that should help with this migration: https://docs.buf.build/migration-prototool/. I could potentially do this myself if we agree it's a good idea.

johanbrandhorst (approved):

moadz (approved):
LGTM

MarcusLongmuir:
Thanks for the reviews, both.

@johanbrandhorst:
> Consider migrating Go protobufs to APIv2

I'll need to see what implications this has for us and then what it looks like for this project implementation and documentation-wise.

> migrate protobuf generation and linting from `prototool` to `buf`

I'd be happy to see this happen. It's only the integration test and example projects that would be affected so it's not really an external change and it would clean up some of the scripts and make things a bit more maintainable.

Issue #899, opened 2021-03-08
Cancelation event not cached by server


watjurk:
### **Versions of relevant software used:**
@improbable-eng/grpc-web: v0.14.0 (go)
@improbable-eng/grpc-web: 0.14.0 (js)

### **What happened:**
Transport: Websocket.
When user makes `unary` or `server stream` request and cancels it, the cancelation event 
isn't cached by the server.

### **What you expected to happen:**
Server should cach the cancelation event.

### **How to reproduce it (as minimally and precisely as possible):**
Transport: Websocket.
Create `unary` or `server stream` service, cancel it on user side and try caching this cancel event on server.

### **My fix:**
webSocketWrappedReader in websocket_wrapper.go  should read one message ahead from websocket.
I've already implemented this in code, so if you would like to I could make a PR.

johanbrandhorst:
What do you mean by caching the cancellation exactly? We have tests that test cancellation propagation.

watjurk:
I meant receiving on context.Cancel(). Your test may not catch this because this error occurs only when transport is set to websocket.

watjurk:
Hi, so what are your thoughts on this ? If you would like to I can provide example repo with this issue.

johanbrandhorst:
The best solution would be in the form of a PR with a failing test case. You can find the cancellation tests here: https://github.com/improbable-eng/grpc-web/blob/master/integration_test/ts/src/cancellation.spec.ts

watjurk:
I'm in progress of creating those test but when I try to run them I'm getting an error: ```Error: This environment's XHR implementation cannot support binary transfer```. This error repeats for every test case.

johanbrandhorst:
Hm, I don't have much experience with the tests, but are you running it in node by any chance? Or, like, a really old browser?

watjurk:
Good question, I assume they are running in node, but I expected this to work because I'm running `npm run start` in `integration_test` folder.

watjurk:
I think I've resolved my issue. Test are running successfully now.

johanbrandhorst:
Do you mean that the original issue is no longer present? If so could you please close the issue?

watjurk:
No the original issue is still present, I mean that I know how to run your's tests successfully.

watjurk:
Okay, I've just modified `cancellation.spec.ts` to `runWithSupportedTransports` and test is failing with websocket transport

johanbrandhorst:
[...]

PR #904, opened 2021-03-10
Abort propagation with websocket transport.

watjurk:
## Changes
Test `should handle aborting a streaming response mid-stream with propagation of the disconnection to the server` in `cancellation.spec.ts` now runs for every supported transport.

Related: #899


johanbrandhorst (approved):
This LGTM but this is @MarcusLongmuir domain of expertise so I will defer the final decision to him.

MarcusLongmuir (approved):

watjurk:
For your information this PR does not contain fix for this test case. CI is passing all tests because of problem see: https://github.com/improbable-eng/grpc-web/issues/899#issuecomment-797385600.

MarcusLongmuir:
Thanks for clarifying. I misunderstood the intent of this PR then. I'll look into the CI problem now.

PR #917, opened 2021-03-16
Fix Websocket Cancellation Handling

### LINKED ISSUE #899: Cancelation event not cached by server

MarcusLongmuir:
## Changes

* Websocket needs to continue reading to capture cancel event
* Fixes #899 
* Re-commits #904 after fixing CI in #909 
* Added explicit websocket cancellation test

## Verification

* Tested locally using Chrome 89 and will check CI thoroughly

### DIFF
--- a/go/grpcweb/websocket_wrapper.go
+++ b/go/grpcweb/websocket_wrapper.go
@@ -176,6 +176,16 @@ func (w *webSocketWrappedReader) Read(p []byte) (int, error) {
 
 	// If the frame consists of only a single byte of value 1 then this indicates the client has finished sending
 	if len(framePayload) == 1 && framePayload[0] == 1 {
+		go func() {
+			for {
+				messageType, _, err := w.wsConn.Read(w.context)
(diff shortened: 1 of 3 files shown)


johanbrandhorst (approved):
Brilliant Marcus :)

Issue #897, opened 2021-03-05
How to use request type GET instead of POST?


pvo13:
Is there a way to specify the request type for the calls grpc-web does? I only see POST requests, I would like to use GET.

I want to use GET requests because it would be a lot easier to cache the results in a PWA service worker since there are solutions for that readily available.

johanbrandhorst:
The gRPC-Web protocol mandates POST requests for all requests. See https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md (which references the HTTP/2 spec: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests).

pcj:
The spec actually does not specify behavior over HTTP 1.1 (did I miss it?).  There is no specific mention of `POST` in the first document.  The gRPC HTTP/2 spec, is IMHO completely orthogonal to grpc-web (otherwise grpc-web would not be needed in the first place!).

Personally I think `GET` should be supported on a per-request opt-in basis for exactly the reason that @pvo13 mentioned: service workers.  In order to cache a response in a service worker for a POST, you have to intercept it and put it in something like IndexedDB.

johanbrandhorst:
The second paragraph in [the spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md#grpc-web) is:

> This document lists the differences between the two protocols. To help tracking future revisions, this document describes a delta with the protocol details specified in the native gRPC protocol.

Thus, because the native gRPC protocol mandates POST, so does grpc-web.

pcj:
OK you are right I did not see the part about the delta.

Nevertheless, I tried it and it works fine. 

On the client side, rather than putting the framed binary request content as the xhr/fetch body, I base64-encoded it and put it into a new header `x-grpc-web-request` and used method `GET` (implies an ~8KB request size limit if what I read about header size limits are true).

On the server, it's a fairly minimal change:

```diff
// IsGrpcWebRequest determines if a request is a gRPC-Web request by checking that the "content-type" is
// "application/grpc-web" and that the method is POST.
func (w *WrappedGrpcServer) IsGrpcWebRequest(req *http.Request) bool {
-	return req.Method == http.MethodPost && strings.HasPrefix(req.Header.Get("content-type"), grpcWebContentType)
+	return strings.HasPrefix(req.Header.Get("content-type"), grpcWebContentType)
}


[...]

PR #935, opened 2021-04-02
Add experimental enable method GET option

pcj:
This experimental PR adds support for handling `GET` requests on the server.  See https://github.com/improbable-eng/grpc-web/issues/897 for context.

### DIFF
--- a/go/grpcweb/options.go
+++ b/go/grpcweb/options.go
@@ -22,6 +22,7 @@ type options struct {
 	corsForRegisteredEndpointsOnly bool
 	originFunc                     func(origin string) bool
 	enableWebsockets               bool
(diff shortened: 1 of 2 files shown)


johanbrandhorst:
I don't think the use case here is strong enough for us to merge this, at least not until there's some official word on the likelihood of this becoming part of the standard. It seems to me the cost of not merging this now is that PWAs cannot rely on GET cache semantics. Is that it?

pcj:
Right.  Currently grpc-web endpoint responses can't be cached by a service worker without something like this.

I'm not exactly expecting this to be merged at this time, I just put it up in case others want to try the code.  Feel free to close if you like.

PR #942, opened 2021-04-14
Add return statement to example

AaronReboot:
As stated, the example would have grpc-web requests execute twice. Adding return to exit handler once grpc-web request is executed


johanbrandhorst (approved):
LGTM

PR #1023, opened 2021-07-08
Add configuration support for nyhoor websocket read limit

bryanmcgrane:
Adding support to configure the underlying nyhoor websocket read limit.

## Changes

The changes are minimal

1. Add new configuration property `WithWebsocketsMessageReadLimit`
2. Apply this read limit to the websocket if it is configured, otherwise do nothing. In this case the default limit is used.

## Verification

The codes successfully builds and works in my local environment.

### DIFF
--- a/go/grpcweb/options.go
+++ b/go/grpcweb/options.go
@@ -24,6 +24,7 @@ type options struct {
 	enableWebsockets               bool
 	websocketPingInterval          time.Duration
 	websocketOriginFunc            func(req *http.Request) bool
+	websocketReadLimit             int64
 	allowNonRootResources          bool
 	endpointsFunc                  *func() []string
 }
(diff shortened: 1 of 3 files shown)


bryanmcgrane:
FYI - the build failed due to changes outside of my PR. It looks like the go.mod file has some linting issues.

johanbrandhorst (changes requested):
Hi, thanks for this change! I think it's missing an additional flag for the grpcwebproxy standalone proxy that could configure this setting for the standalone proxy. What do you think?

bryanmcgrane:
Good thinking. That change has been made as well!

johanbrandhorst:
Looks like CI is failing because of some module inconsistencies. Could you run `go mod tidy` and commit the changes?

johanbrandhorst (approved):
Scratch that, I don't see why this change should cause any import changes.

johanbrandhorst:
Thanks for your contribution!

bryanmcgrane:
You're welcome1

PR #1048, opened 2021-08-03
Fixed go.sum and go.mod for CI

MarcusLongmuir:
## Changes

* Short-term fix for the `git diff` breakage in CI that is blocking PRs by committing the updated `go.mod` and `go.sum`


johanbrandhorst (approved):
Is the fact that CI is failing even after this fix concerning?

johanbrandhorst (approved):

PR #1061, opened 2021-11-03
Disable websocket compression

jsmouret:
It seems that iOS 15 does not support permessage-deflate so workaround it by disabling compression.

A better PR would properly detect if the extension is supported, and maybe patch [nhooyr.io/websocket](https://github.com/nhooyr/websocket) instead.

For people in production, edit go.mod and add:
```
replace github.com/improbable-eng/grpc-web v0.14.1 => github.com/jsmouret/grpc-web v0.14.2-0.20211103063242-8c932b2237aa
```


jsmouret:
Still useful

PR #1066, opened 2022-01-18
updated documentation to accurately reflect the existing API for unary call aborting

seanwatters:
The current documentation states that `.cancel()` is the method for aborting a unary call—it appears as though both `unary` and `invoke` actually use `.close()`.

## Changes

This PR updates the documentation to accurately reflect the API in source code.

## Verification

Viewing source code and using TypeScript intellisense.

### DIFF
--- a/client/grpc-web/README.md
+++ b/client/grpc-web/README.md
@@ -45,11 +45,11 @@ grpc.unary(BookService.GetBook, {
 });
 ```
 
-* Requests can be aborted/cancelled before they complete:
+* Requests can be aborted/closed before they complete:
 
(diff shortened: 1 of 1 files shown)


johanbrandhorst (approved):
Looks reasonable to me https://github.com/improbable-eng/grpc-web/blob/80f55434014877ea8716fbab95618cc13b2e9572/client/grpc-web/src/client.ts#L318-L329.

johanbrandhorst:
CI is broken, but clearly this docs change didn't cause it, so I'll merge it.

PR #1068, opened 2022-01-23
feat(client): remove peer depedency of google-protobuf

HenriBeck:
In the main grpc-web package, the google-protobuf package is declared as a peer dependency even though it is not referenced anywhere in the source code and doesn't need to be installed to use grpc-web.

## Changes

Removed the `google-protobuf` package as a dependency.

## Verification

I tried running the integration tests but I didn't manage to even start them on my machine. The change however should pose no problems because only a peer dependency was removed.


johanbrandhorst:
This sounds reasonable to me, unfortunately I'm not well versed enough in the TypeScript side of this to just merge when the CI is broken like this. CC @MarcusLongmuir for review.

PR #1070, opened 2022-01-28
Fix go mod tidy

johanbrandhorst:


johanbrandhorst:
@MarcusLongmuir this should fix the broken CI runs in other PRs. We're still struggling with a max of 5 concurrent saucelabs tests.

MarcusLongmuir (approved):

Issue #1081, opened 2022-02-13
[Feature] multiplex over a websocket


borissmidt:
Dear,

I was wondering, could a websocket be multiplexed so it really mirrors the java/go api? So you create 1 connection per stub, this would allow grpc-web to have a huge amount of streams.

Maybe a simple protocol could be defined in protobuf itself to make it easy to implement in other languages to have an InProcessProxy reducing the need for proxies.

The only disadvantage is that the proxy then has to keep state to track which message maps to which stream.

Would you be interested in such a feature or would it be too complex to implement? 

For the protocol i was thinking along the lines of: 

```proto
syntax = "proto3";

option go_package = "./main";

message Message{
    uint32 streamId = 1;
    oneof payload {
        Start start = 2;
        Body body = 3;
        Complete complete = 4;
        Error error = 5;
        Cancel cancel = 6;
    } 
}

message Start {
    string operation = 1;
    repeated Header headers = 2;
}

message Body {
    bytes data = 1;
}

message Complete {
}

message Error {
}


message Cancel {
}

message Header {
    string name = 1;
    bytes data = 2;
}
```

borissmidt:
how would that work then on the go side? would you get a special websocket implementation that then gives back somekind of 'virtual' websocket?

johanbrandhorst:
That I do not know

borissmidt:
Would it be technically feasible? (i was trying implement it but i have never written go before :smile: )

johanbrandhorst:
Oh I'm sure it would be, but the websocket abstraction on the backend would need to perform all the multiplexing work unless the standard library can somehow abstract that. https://datatracker.ietf.org/doc/html/rfc8441 this appears to be the RFC. https://github.com/nhooyr/websocket/issues/4 is the issue for our websocket library, which seems to depend on https://github.com/golang/go/issues/32763.

borissmidt:
I will try to implement it in scala, i'm more experienced with that, and if it was easy to implement i will port it to go and the typescript client.

borissmidt:
I tried to implement it in go but i get an 
```INFO[0000] streamId:1 header:{headers:{key:"Content-Type" value:"text/plain; charset=utf-8"} headers:{key:"X-Content-Type-Options" value:"nosniff"} status:500} ```
so i can get the headers back to the client which is already a great achievement.
https://github.com/borissmidt/grpc-web-1/blob/webscocketChannel/go/test/server_test.go

[...]

PR #1084, opened 2022-02-19
[Work in progress] Webscocket channel, Do not merge.

borissmidt:
## Changes
adds support to use a websocket like a managed channel/connection and reuse it for all requests. 
#1081

Todo:
- [x] Add websocketChannel to the go proxy
- [ ] Test the websocketChannel proxy, to verify that there is no resource leak.
- [ ] Backpressure between proxy and client (nice to have)
- [ ] Java proxy implementation (probably for another repo)
- [ ] Add integration tests.
- [x] Typescript: Add websocketChannel to the typescript client.
- [x] Typescript: Channel recovery. (how is this done in http2)
- [ ] Cleanup commits
- [x] max number of streams setting per websocket channel
- [ ] ???

The code currently has been tested at [this commit](https://github.com/improbable-eng/grpc-web/commit/9acf8244d00973529b8c9d54b3f6cc97f71c9753). More low level tests have to be added.


go/grpcwebproxy/main.go
```diff
@@ -44,6 +44,8 @@ var (
 	websocketPingInterval = pflag.Duration("websocket_ping_interval", 0, "whether to use websocket keepalive pinging. Only used when using websockets. Configured interval must be >= 1s.")
 	websocketReadLimit    = pflag.Int64("websocket_read_limit", 0, "sets the maximum message read limit on the underlying websocket. The default message read limit is 32769 bytes.")
 
+	useWebsocketsChannel  = pflag.Bool("use_websockets_channels", false, "whether to use a chanllel over websocket transport layer (alpha)")
```
hexfusion, line 47: minor: s/chanllel/channel

go/grpcweb/wrapper.go
```diff
@@ -221,6 +236,40 @@ func (w *WrappedGrpcServer) IsGrpcWebRequest(req *http.Request) bool {
...
+	headers := make(http.Header)
+	for _, name := range w.allowedHeaders {
+		if values, exist := req.Header[name]; exist {
+			headers[name] = values
+		}
+	}
+
+	ctx, cancelFunc := context.WithCancel(req.Context())
+	defer cancelFunc()
+	NewWebsocketChannel(wsConn, w.handler, ctx)
```
hexfusion, line 264: we need to show were this went :)

go/grpcweb/wrapper.go
```diff
@@ -123,6 +123,15 @@ func (w *WrappedGrpcServer) ServeHTTP(resp http.ResponseWriter, req *http.Reques
 		resp.WriteHeader(http.StatusForbidden)
 		_, _ = resp.Write(make([]byte, 0))
 		return
+	} else if w.opts.enableWebsocketChannels && w.IsGrpcWebSocketChannelRequest(req) {
+		if w.websocketOriginFunc(req) {
+			if !w.opts.corsForRegisteredEndpointsOnly || w.isRequestForRegisteredEndpoint(req) {
+				w.HandleGrpcWebsocketRequest(resp, req)
```
hexfusion, line 129: This should be `HandleGrpcWebsocketChannelRequest` right?

hexfusion (commented):
Added a few notes on small nits I know you're this working n this. Going to implement this and test it tonight/tomorrow.

go/test/server_test.go
```diff
@@ -0,0 +1,240 @@
...
+
+	serverResponse = &HelloReply{}
+	if err := proto.Unmarshal(response2.GetBody().GetData(), serverResponse); err != nil {
+		fmt.Errorf("error %v", err)
+	}
+	log.Infof("got server response2 %v", serverResponse)
+	c.Close(websocket.StatusNormalClosure, "")
+}
+
+func CompleteGrpc(ws *websocket.Conn, streamId uint32, ctx context.Context) error {
```
hexfusion, line 149: How is the stream id managed? I might of missed that.

borissmidt: The client will give streamid's back and has to manage it properly.
I would use an incremental id and assume that you don't do 4g requests. 
A max number of streams per channel might be useful as protection.

go/grpcweb/wrapper.go
borissmidt: Good catch.

borissmidt:
It seems that yesterday i forgot to commit a lot of files, i now added the typescript code but i sitl have to add the integration test.

borissmidt:
I really get stuck at the integration tests, i want to analyse what goes wrong, but there are no nice stack traces because of the webpack build.
I also have a lot of trouble to keep the build consistently working when i make some changes to the libary and then want to rerun the tests. (`npm link ../client/web-grpc` doesn't seem to work properly.)

I think the channel never opens and that is why the tests fail but i cannot debug it which makes the further development very hard.

johanbrandhorst:
Hi Boris,

Sorry that you went through all this work, but I honestly don't think we can accept this large of a contribution to the code base at this stage. This code is mostly in maintenance mode and I'm not confident that we'd be able to maintain this new addition. It appears you're run into some trouble with our tests, and I'm sorry about that, but unfortunately they're one of the problems with the current code base.

borissmidt:
Ah to bad, atleast i learned some typescript and go on these rainy days. 
I still believe that this would solve some of the grpc-web problems and increase the support on the browsers.
Maybe i better start a fork and name it 'grpc-websocket' because i'm 100% sure that grpc/grpc-web are not open to websocket implementations.

borissmidt:
The last test run i had, it seems that it works:
[logging.log](https://github.com/improbable-eng/grpc-web/files/8112421/logging.log)

johanbrandhorst:
You are of course welcome to fork the project and do your own development as long as you abide by the license: https://github.com/improbable-eng/grpc-web/blob/master/LICENSE.txt

johanbrandhorst:
I'm going to close this PR, since we have no intention of landing it.

PR #1092, opened 2022-03-14
Limit parallelism of browser tests

johanbrandhorst:
Sauce labs only allow 5 tests to run concurrently, so limit parallelism to avoid having to manually rerun tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,6 +62,7 @@ jobs:
       browser-params:
         type: string
     executor: test-executor
+    parallelism: 5
     steps:
       - checkout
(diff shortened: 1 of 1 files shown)


johanbrandhorst:
image

🎉

johanbrandhorst:
This seems to be failing for unrelated reasons. @MarcusLongmuir @easyCZ I think we should merge this as it's doing what I intended, which is limiting the saucelab tests to 5 at a time. I don't know what the reason for the test failure is.

johanbrandhorst:
Looks like it succeeded on rerun. Please review @MarcusLongmuir @easyCZ so we can merge this.

MarcusLongmuir (approved):

PR #1112, opened 2022-07-08
Server websocket implementation should allow, ignore other subprotocols

### LINKED ISSUE #1111: grpcwebproxy websocket implementation fails on multiple subprotocols
At least 0.12 through latest

**What happened**
Websockets support a mechanism known as "subprotocols" where a connecting client can offer zero to many options, and the server can pick from that list the best suited. The server then should communicate that negotiated protocol back to the client so both sides know which protocol is in use for the duration of the connection as the websocket finishes opening, and both client and server then can proceed sending messages using that protocol.

The websocket spec calls for the `Sec-Websocket-Protocol` header(s) to be defined by the client (multiple headers can be set, each header entry should be comma separated). The server then must select one of these values, or respond that it does not accept these options. See https://datatracker.ietf.org/doc/html/rfc6455#section-4.1, See https://datatracker.ietf.org/doc/html/rfc6455#section-4.2 for more details.

The use case here is to implement an alternative but compatible websocket client/server (see #1081, #198) that is able to share a single websocket for all active streams. The proposed client implementation would support both `grpc-websockets` and `grpc-websockets-multiplex`. Server negotiation should inform the client which implementation will be used.

Presently, grpcwebproxy's websocket implementation when offered multiple protocols will simply fail the connection, rather than pick the acceptable option `grpc-websockets`. 

If the client websocket contains `grpc-websockets` as an option, it is expected that the server will respond by opening the websocket selecting that as the negotiated subprotocol. Note that the `nhooyr.io/websocket` library correctly handles this, but only the `IsGrpcWebSocketRequest` check incorrectly fails.

The simplest way is to modify client/grpc-web/src/transports/websocket/websocket.ts to also offer any other protocol along with `grpc-websockets`, under the assumption that any server which can use that protocol will respond back with that same string. For example:
```
    start: (metadata: Metadata) => {
      ws = new WebSocket(webSocketAddress, ["grpc-websockets"]);
      ws.binaryType = "arraybuffer";
```

I'll offer a pull request to fix this. The project github.com/deephaven/deephaven-core has an in-process grpc-web and grpc-websockets proxy, and is seeking to add a grpc-websockets-multiplex proxy for faster non-tls stream initiation (and to avoid exhausting the browser's allowed websocket count). If there is interest, we can provide a ts client here, and see about an update to the Go proxy provided by improbable-eng/grpc-web.

niloc132:
Correctly reads all Sec-Websocket-Protocol headers and reads all values
in those headers, and only succeeds if the expected subprotcol value of
"grpc-websockets" is present. This implementation roughly mirrors what
the nhooyr.io/websocket Accept() method does, with a few
simplifications.

Fixes #1111

## Changes

Updated `IsGrpcWebSocketRequest()` to allow for multiple protocol headers and values, but still only succeeds if the expected value `grpc-websockets` is present. Uses `strings.EqualFold` in place of `strings.toLower` and a `==` check, as this is a more flexible check, and reflects the underlying nhooyr.io/websockets implementation.

## Verification

* Tested with a lightly modified websocket.ts transport, which advertises an alternative subprotocol.
* Tested with a multiplex-capable websocket transport, which can fall back to the existing websocket transport in case the server doesn't also support multiplex.

### DIFF
--- a/go/grpcweb/wrapper.go
+++ b/go/grpcweb/wrapper.go
@@ -135,19 +135,19 @@ func (w *WrappedGrpcServer) ServeHTTP(resp http.ResponseWriter, req *http.Reques
 // IsGrpcWebSocketRequest determines if a request is a gRPC-Web request by checking that the "Sec-Websocket-Protocol"
 // header value contains "grpc-websockets"
 func (w *WrappedGrpcServer) IsGrpcWebSocketRequest(req *http.Request) bool {
-    if strings.ToLower(req.Header.Get("Upgrade")) != "websocket" {
-        return false
-    }
+	if strings.ToLower(req.Header.Get("Upgrade")) != "websocket" {
+		return false
+	}
 
-    for _, subproto := range req.Header["Sec-Websocket-Protocol"] {
-        for _, token := range strings.Split(subproto, ",") {
-            token = strings.TrimSpace(token)
-            if strings.EqualFold(token, "grpc-websockets") {
-                return true
-            }
-        }
-    }
-    return false
+	for _, subproto := range req.Header["Sec-Websocket-Protocol"] {
+		for _, token := range strings.Split(subproto, ",") {
+			token = strings.TrimSpace(token)
+			if strings.EqualFold(token, "grpc-websockets") {
+				return true
+			}
+		}
+	}
+	return false
 }
 
 // HandleGrpcWebRequest takes a HTTP request that is assumed to be a gRPC-Web request and wraps it with a compatibility
--- a/go/grpcweb/wrapper.go
+++ b/go/grpcweb/wrapper.go
@@ -139,7 +139,7 @@ func (w *WrappedGrpcServer) IsGrpcWebSocketRequest(req *http.Request) bool {
 		return false
 	}
 
-	for _, subproto := range req.Header["Sec-Websocket-Protocol"] {
+	for _, subproto := range req.Header.Values("Sec-Websocket-Protocol") {
 		for _, token := range strings.Split(subproto, ",") {
 			token = strings.TrimSpace(token)
 			if strings.EqualFold(token, "grpc-websockets") {
--- a/go/grpcweb/wrapper.go
+++ b/go/grpcweb/wrapper.go
@@ -133,9 +133,21 @@ func (w *WrappedGrpcServer) ServeHTTP(resp http.ResponseWriter, req *http.Reques
 }
 
 // IsGrpcWebSocketRequest determines if a request is a gRPC-Web request by checking that the "Sec-Websocket-Protocol"
-// header value is "grpc-websockets"
+// header value contains "grpc-websockets"
 func (w *WrappedGrpcServer) IsGrpcWebSocketRequest(req *http.Request) bool {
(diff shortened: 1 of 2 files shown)


go/grpcweb/wrapper.go
```diff
@@ -133,9 +133,21 @@ func (w *WrappedGrpcServer) ServeHTTP(resp http.ResponseWriter, req *http.Reques
...
 // IsGrpcWebSocketRequest determines if a request is a gRPC-Web request by checking that the "Sec-Websocket-Protocol"
-// header value is "grpc-websockets"
+// header value contains "grpc-websockets"
 func (w *WrappedGrpcServer) IsGrpcWebSocketRequest(req *http.Request) bool {
-	return strings.ToLower(req.Header.Get("Upgrade")) == "websocket" && strings.ToLower(req.Header.Get("Sec-Websocket-Protocol")) == "grpc-websockets"
+	if strings.ToLower(req.Header.Get("Upgrade")) != "websocket" {
+		return false
+	}
+
+	for _, subproto := range req.Header["Sec-Websocket-Protocol"] {
```
johanbrandhorst, line 142: Slightly safer than this, can we use [Values](https://pkg.go.dev/net/http#Header.Values)? It performs canonicalization of the header.
```suggestion
	for _, subproto := range req.Header.Values("Sec-Websocket-Protocol") {
```

johanbrandhorst (changes requested):
One very minor nit, this looks great.

niloc132:
The "test failure" looks like it might be a transient saucelabs tunnel issue?

https://app.circleci.com/pipelines/github/improbable-eng/grpc-web/665/workflows/1ccb9f11-3ca4-403e-b584-40379472b9de/jobs/6077

```
...
+ sleep 1
+ test 2 -eq 0 -o -e ./sauce-connect-readyfile-no-ssl-bump
+ sleep 1
+ test 1 -eq 0 -o -e ./sauce-connect-readyfile-no-ssl-bump
+ sleep 1
+ test 0 -eq 0 -o -e ./sauce-connect-readyfile-no-ssl-bump
+ (( ++wait_seconds ))
+ echo 'Timed out waiting for sauce labs tunnel (with ssl bump)'
+ kill 1369
./run-with-tunnel.sh: line 93: kill: (1369) - No such process
+ killTunnels
+ echo 'Killing Sauce Labs Tunnels...'
62.085767+ [[ -z '' ]]
 Timed out waiting for sauce labs tunnel (with ssl bump)
+ kill 1369
./run-with-tunnel.sh: line 79: kill: (1369) - No such process
62.085857 Killing Sauce Labs Tunnels...
npm ERR! code 1
npm ERR! path /go/src/github.com/improbable-eng/grpc-web/integration_test
npm ERR! command failed
npm ERR! command sh -c ./run-with-tunnel.sh ./run-with-testserver.sh karma start ./karma.conf.ts --single-run
...
```

Can someone re-run that?

johanbrandhorst (approved):

johanbrandhorst:
Thank you for your contribution!

niloc132:
@johanbrandhorst would it be reasonable to request a release be cut, for this and a few other fixes made since the release last November?

johanbrandhorst:
That sounds reasonable :). The release process isn't super automated, so I'm gonna have to ask @MarcusLongmuir for some help.

johanbrandhorst:
Ping @MarcusLongmuir or @easyCZ, either of you have bandwidth for a new release?

PR #1116, opened 2022-07-22
Add websocket compression mode option

yurifedoseev:
## Changes

1. Add WebsocketCompressionMode option to WrappedGrpcServer for managing websocket compression
2. Add usage example to grpcwebproxy

## Verification
The motivation: we encountered some websocket problems with IOS 15 which were fixed by disabling compressing.
I'd like to have an option for managing websocket compression depending on the application needs.
I've tested disabling websocket compression in a separate fork

### DIFF
--- a/go/grpcweb/options.go
+++ b/go/grpcweb/options.go
@@ -5,8 +5,9 @@ package grpcweb
 
 import (
 	"net/http"
-	"nhooyr.io/websocket"
 	"time"
+
+	"nhooyr.io/websocket"
 )
 
 var (
--- a/go/grpcweb/options.go
+++ b/go/grpcweb/options.go
@@ -5,6 +5,7 @@ package grpcweb
 
 import (
 	"net/http"
+	"nhooyr.io/websocket"
 	"time"
 )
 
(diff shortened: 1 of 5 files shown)


yurifedoseev:
I believe I have some problems with godocdown. I've used the latest version from github

niloc132 (commented):
Perhaps also add docs for this new flag to https://github.com/improbable-eng/grpc-web/blob/master/go/grpcwebproxy/README.md?

go/grpcweb/options.go
```diff
@@ -15,6 +17,7 @@ var (
 		originFunc:                     func(origin string) bool { return false },
 		allowNonRootResources:          false,
 		corsMaxAge:                     10 * time.Minute,
+		websocketCompressionMode:       websocket.CompressionNoContextTakeover,
```
johanbrandhorst, line 20: Is this the current value?

johanbrandhorst (changes requested):
This sounds reasonable to me, but as suggested by @niloc132, could you add a section to the README?

go/grpcweb/options.go
yurifedoseev: Yes, here is the definition https://github.com/nhooyr/websocket/blob/master/compress.go#L22
`iota` means starting from 0

yurifedoseev:
Updated the README.md

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution!

PR #1120, opened 2022-08-18
Add documentation to show client-side/bi-directional streaming using websockets

shreyas-samsung:
## Changes

The main and client/grpc-web Readme files do not contain info on the usage of websocket transport for supporting client-side or bi-directional streaming RPCs.

1. This info has been added in both the Readme files.

2. A new doc is created that explains how to configure websockets at the client and server side.


johanbrandhorst (approved):
This looks great, thanks!

PR #1153, opened 2023-02-09
Notify users that this repo is in maintenance mode

louisrli:
Hey there, I've recently been trying to figure out the open source proto infrastructure, which in itself has been pretty hard to navigate -- but I noticed that @johanbrandhorst has commented on multiple recent issues with this quote.

> Please note that this repo is in maintenance mode, and we recommend users migrate to the official grpc-web client: https://github.com/grpc/grpc-web.

It's interesting that it doesn't seem to suggest this anywhere on the README, and I didn't really figure out that this was the suggested path (by the maintainers of this repo) until digging deeper while debugging some stuff. In the end, it's totally up to you, but it seems to me like this would be something useful to put at the top of the README, so this PR just copy-pastes that quote (feel free to adjust formatting / wording as desired).

Cheers, and thanks for your work.


johanbrandhorst (approved):
Thanks for this, you're totally right.

a-priestley:
I think there should be a caveat to this recommendation in that the official client does not yet support binary streams. I switched to their client and ended up switching back because processing for my project was too slow over grpc-web-text. I intend to stick with this client until theres implements it.

johanbrandhorst:
I'd recommend switching to Connect nowadays. https://connect.build

a-priestley:
> I'd recommend switching to Connect nowadays. https://connect.build

Thank you for the suggestion -- I'll check it out!

